PR #10, opened 2019-11-19
Update to point the HAFS related developments for the write_grid_component in fv3atm

BinLiu-NOAA:
This Pull Request is to point the proper commit for the FV3 submodule (fv3atm), which has the HAFS related developments for the write_grid_component to
- Enable outputting a grid (regional_latlon, rotated_latlon, or lambert_conformal) larger than the native computation grid for the regional or nested domain/tile.
- Enable writing out the output on the native computation grid directly without any interpolation for the regional or nested domain/tile. 
Also the missing value in physics bundle is changed from real(kind_phys) to real in io/FV3GFS_io.F90 as the file could be compiled in 32bit real or 64bit real.


BinLiu-NOAA:
Notes: 
- Currently, it points to the support/HAFS branch in the hafs-community fork of fv3atm. After the regression tests and the develop branch of fv3atm being updated, will change to point to the proper commit in the develop branch of the fv3atm repo.
- Regression tests have been done on Hera, confirming these write_grid_component changes do not change the baseline results.

DusanJovic-NOAA (approved):

junwang-noaa (approved):

PR #11, opened 2019-11-25
Add support for cmake build system.

DusanJovic-NOAA:


cmake/FindESMF.cmake
```diff
@@ -0,0 +1,47 @@
+
```
climbfuji, line 1: I wonder if this could be fed back to the cmake repos so that findpackage(ESMF) will be available to all users.

cmake/GNU.cmake
```diff
@@ -0,0 +1,58 @@
...
+else()
+    message("DEBUG is       disabled (optimized build)")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -O2 -fno-range-check")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2")
+endif()
+
+
+if(REPRO)
+    message("REPRO is       ENABLED")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -O2 -ggdb")
```
climbfuji: I found that on macOS, you need to change `-O2` to `-O0` to get b4b identical results between IPD and CCPP if (and only if) the dynamics are compiled in 32bit (independent of whether gcc+gfortran or clang+gfortran are used). I would suggest to have another test inside the `if(REPRO)` block that tests for macOS (I believe `if(APPLE)` works) that uses `-O0` instead of `-O2`. For Linux, this problem doesn't exist.

cmake/GNU.cmake
```diff
@@ -0,0 +1,58 @@
...
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2")
+else()
+    message("REPRO is       disabled")
+endif()
+
```
climbfuji, line 27: I would also add a mechanism to either abort if both REPRO and DEBUG are requested, or make one of the two superior (I suggest DEBUG). In the latter case, issue a warning that both REPRO and DEBUG where requested, and that DEBUG will be used (and REPRO disabled).

cmake/GNU.cmake
```diff
@@ -0,0 +1,58 @@
...
+    message("OPENMP is      ENABLED")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fopenmp")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fopenmp")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fopenmp")
+    add_definitions(-DOPENMP)
+else()
+    message("OPENMP is      disabled")
+endif()
+
+if(AVX2)
```
climbfuji, line 49: GNU doesn't have AVX2 flags, i.e. here it would be better to abort and tell the user to do his/her homework.

cmake/PGI.cmake
```diff
@@ -0,0 +1,62 @@
+
```
climbfuji, line 1: Is anyone going to test this? I won't.

cmake/configure_linux.gnu.cmake
```diff
@@ -0,0 +1,35 @@
+message("")
```
climbfuji, line 1: We will need a cmake/configure_macos.gnu.cmake as well.

modulefiles/wcoss_dell_p3/fv3
```diff
@@ -17,6 +17,7 @@ module-whatis "loads NEMS FV3 prerequisites on Venus and Mars"
 ##
```
climbfuji, line 17: I guess we need to add these CMAKE_... stanzas for linux.gnu, linux.intel and macos.gnu as well?

tests/rt_utils.sh
```diff
@@ -442,49 +444,52 @@ rocoto_create_compile_task() {
...
-    <cores>1</cores>
-    <walltime>01:00:00</walltime>
-    <join>&LOG;/compile_${COMPILE_NR}.log</join>
-    ${NATIVE}
-  </task>
-EOF
-  else
-    cat << EOF >> $ROCOTO_XML
-  <task name="compile_${COMPILE_NR}" maxtries="1">
+  # serialize CCPP builds. FIXME
```
climbfuji: This should be ok in parallel now that we have the code in ccpp-physics and ccpp-framework. I can try to add the missing bits after this PR has been merged.

climbfuji (commented):
Looks great, a lot of work. I suppose I will be doing the macos and cheyenne parts after this PR has been merged?

cmake/PGI.cmake
DusanJovic-NOAA: Probably not. Should we remove all references to PGI?

climbfuji: I just read through the platform and compiler support policy. PGI is listed as Tier 3, and per definition for Tier 3 compiler option template must be provided (albeit compilation and execution are not guaranteed). So, let's keep it untested ...

modulefiles/wcoss_dell_p3/fv3
DusanJovic-NOAA: Since we can not assume that on generic linux machine (ie. laptop/desktop) modules are available, nor location of the libraries or compiler (and mpi wrapper) names, I'm not loading (sourcing) these files. Build scripts (build.sh) sets up all required environment variables.

cmake/GNU.cmake
DusanJovic-NOAA: I left this in case we decide to add options that turn on AVX2, like -march=cpu-type.

cmake/GNU.cmake
DusanJovic-NOAA: I can not test on macOS. Do you want to make required changes?

modulefiles/wcoss_dell_p3/fv3
climbfuji: Got it - makes total sense. Thanks for the clarification.

cmake/GNU.cmake
climbfuji: As long as users don't expect to get faster code by turning that on and nothing happens, that's fine with me.

cmake/GNU.cmake
climbfuji: Will do, yes.

climbfuji:
@DusanJovic-NOAA just a quick question - seems like my original modifications of your cmake_rt branches are not included. Can you confirm?

DusanJovic-NOAA:
> @DusanJovic-NOAA just a quick question - seems like my original modifications of your cmake_rt branches are not included. Can you confirm?

Which modifications? Please remind me.

climbfuji:
> > @DusanJovic-NOAA just a quick question - seems like my original modifications of your cmake_rt branches are not included. Can you confirm?
> 
> Which modifications? Please remind me.

I originally pushed them to gerrit/cmake_rt_gnu_and_ccpp (for NEMSfv3gfs, FV3), and same branches in the GitHub based repositories. But never mind, I got them back in already. Will create a PR later to update your branches.

cmake/FindESMF.cmake
DusanJovic-NOAA: Maybe. But for now we must have it available locally.

tests/rt_utils.sh
DusanJovic-NOAA: Removed outdated comments about serialized ccpp build jobs. See e7dcda0

jedwards4b:
The cime developers at NCAR need this PR merged in order to continue our process of cime integration for the release.  Please note that you do not need any further work for cheyenne build, we will take care of that.

cmake/configure_linux.gnu.cmake
DusanJovic-NOAA: Committed in d62d01a

cmake/GNU.cmake
DusanJovic-NOAA: Committed in d62d01a

DusanJovic-NOAA:
> The cime developers at NCAR need this PR merged in order to continue our process of cime integration for the release. Please note that you do not need any further work for cheyenne build, we will take care of that.

We plan to merge this PR by the end of this week.

climbfuji:
This is all good to go from my end.

junwang-noaa (approved):
The code changes look good to me.

PR #20, opened 2019-12-19
Allow environment variable NEMS_MACHINE to overwrite (or set) MACHINE_ID

climbfuji:
This PR is a follow-up PR to the recent update from DTC. Similar to using the `NEMS_COMPILER` environment variable for setting the compiler when running `rt.sh`, one can use `NEMS_MACHINE` to overwrite (if the logic in `detect.sh` detects a machine) or set (if not) `MACHINE_ID`.

This is (a) required for automatic regression testing on Cheyenne, because `detect.sh` fails to identify the compute nodes as Cheyenne nodes, and (b) generally a good practice to avoid having to add more and more logic to `detect.sh` as we expand the UFS public release regression testing to other machines. 

This is also part of the NCAR dtc/develop branches, has been tested extensively, and will be proposed for inclusion in EMC's develop branch as well.

### DIFF
--- a/tests/detect_machine.sh
+++ b/tests/detect_machine.sh
@@ -122,6 +122,9 @@ case $(hostname -f) in
   login4.stampede2.tacc.utexas.edu) MACHINE_ID=stampede ;; ### stampede4
 esac
 
+# Overwrite auto-detect with NEMS_MACHINE if set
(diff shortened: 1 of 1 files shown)


DusanJovic-NOAA (approved):

climbfuji:
Thanks, Dusan. This was quick.

PR #22, opened 2019-12-27
Model configure defaults

DusanJovic-NOAA:
This PR sets default values for several module_configure variables.
* In NEMS RUN_CONTINUE, ENS_SPS are set to .false. and TOTAL_MEMBER to 1. These variables are not used in public release and can be removed from model_confiure file.
* In FV3 calendar is set to Gregorian to be consistent with the calendar kind used by ESMF.

Associated PRs:
https://github.com/NOAA-EMC/fv3atm/pull/34
https://github.com/NOAA-EMC/NEMS/pull/18


climbfuji (approved):
Looks ok to me! Surprised that the default calendar was still Julian. I wonder if we see any impact of changing to Gregorian when making those changes in the "develop" branch.

DusanJovic-NOAA:
> Looks ok to me! Surprised that the default calendar was still Julian. I wonder if we see any impact of changing to Gregorian when making those changes in the "develop" branch.

@climbfuji I updated baselines only on Hera, since I do not have access to Cheyenne.

climbfuji:
> > Looks ok to me! Surprised that the default calendar was still Julian. I wonder if we see any impact of changing to Gregorian when making those changes in the "develop" branch.
> 
> @climbfuji I updated baselines only on Hera, since I do not have access to Cheyenne.

Sure. In the spirit of cheyenne being a tier-2 platform, I will do this retrospectively.

junwang-noaa (approved):

PR #31, opened 2020-01-16
Updated chapter about model development with material from Wiki

valbonakunkel-noaa:
Ligia,

The chapter about model development is updated with material from Wiki.

Valbona


ligiabernardet (commented):
This is probably not the PR that @valbonakunkel-noaa intended to submit. What we need is a PR from Valbona's fork/branch to the "documentation" branch of the ufs-weather-model. @valbonakunkel-noaa, please create a new PR. Thanks.

PR #33, opened 2020-01-19
Fix restarts and add restart tests for ufs_public_release

climbfuji:
- fix restarts and add restart tests for ufs_public_release
- compile dynamics in 32bit in regression tests
- update stochastic_physics submodule pointer
- to-level build.sh: require COMPILER environment variable only if `CMAKE_Platform` is not set (this allows for using build.sh on one of the platforms intrinsically supported by the ufs-weather-model: `module use $PWD/modulefiles/hera.intel/fv3 && ./build.sh`)

See https://github.com/NOAA-EMC/fv3atm/issues/42 and https://github.com/ufs-community/ufs-mrweather-app/issues/62 for a lengthy discussion of the issue and a description of the solution. Essentially, change `nstf_name` from `2,1,0,0,0` to `2,1,1,0,5` (and flip the second `1` to `0` for restarts) - and turn off the stochastic_physics (`do_sppt`, `do_shum`, `do_skeb`), which should have been the default setting anyway.

New baselines have been generated on Cheyenne for Intel and GNU (both are in place, final round of regression tests against this baseline have been kicked off and logs will be updated shortly). For hera.intel, the new baseline temporarily is in
```
/scratch1/NCEPDEV/stmp4/Dom.Heinzeller/FV3_RT/REGRESSION_TEST_UFS_PUBLIC_RELEASE_INTEL_FOR_BASELINE_20200121
```
and should become
```
/scratch1/NCEPDEV/nems/emc.nemspara/RT/NEMSfv3gfs/ufs-public-release-20200121
```
@DusanJovic-NOAA can you please rsync the directory and let me know when it is done (next week is sufficient, no hurries) so that I can kick off the final round of regression tests? Thanks!


tests/rt.conf
```diff
@@ -2,20 +2,24 @@
...
-COMPILE | CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                           | standard    | cheyenne.gnu   | fv3         |
+COMPILE | 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                   | standard    | hera.intel     | fv3         |
+COMPILE | 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                   | standard    | cheyenne.intel | fv3         |
+COMPILE | 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                   | standard    | cheyenne.gnu   | fv3         |
 
 RUN     | fv3_ccpp_gfs_v15p2                                                                                                             | standard    |                | fv3         |
+RUN     | fv3_ccpp_gfs_v15p2_coldstart                                                                                                   | standard    |                |             |
```
junwang-noaa, line 10: What is the difference between the two tests fv3_ccpp_gfs_v15p2 and fv3_ccpp_gfs_v15p2_coldstart? If just forecast lengths are different, I would suggest to use Dusan's fix to run fv3_ccpp_gfs_v15p2 and fv3_ccpp_gfs_v15p2_restart.

tests/tests/fv3_ccpp_gfs_v15p2
```diff
@@ -38,6 +50,18 @@ export LIST_FILES="atmos_4xdaily.tile1.nc \
                    dynf024.tile4.nc \
                    dynf024.tile5.nc \
                    dynf024.tile6.nc \
+                   dynf027.tile1.nc \
+                   dynf027.tile2.nc \
+                   dynf027.tile3.nc \
+                   dynf027.tile4.nc \
+                   dynf027.tile5.nc \
+                   dynf027.tile6.nc \
```
junwang-noaa, line 58: We may only need compare forecast results at fh048?

tests/tests/fv3_ccpp_gfs_v15p2_coldstart
```diff
@@ -0,0 +1,48 @@
+###############################################################################
+#
+#  FV3 CCPP GFS v15.2 compiled with 32-bit dynamics coldstart (0-24h) test
+#
+###############################################################################
+
+export TEST_DESCR="Compare FV3 32bit CCPP GFS v15.2 results with previous trunk version"
```
junwang-noaa, line 7: I don;t think this test is necessary. It repeats the previous test, we can use the "restart_interval: 24" to get 24 hour restart files from previous test, and it looks to me this method is error-prone as it copied restart files (RESTART/*) without a restart date in the file name.

tests/rt.conf
```diff
@@ -2,20 +2,24 @@
...
+RUN     | fv3_ccpp_gfs_v15p2_restart                                                                                                     | standard    |                |             |
```
DusanJovic-NOAA: This test (fv3_ccpp_gfs_v15p2_restart) must be executed only after the previous one (fv3_ccpp_gfs_v15p2_coldstart) successfully finishes, because it depends on it. Without specifying this dependency explicitly (in the 6th column) parallel execution of regression tests using ecFlow of Rocoto will fail. See how this kind of dependency is specified for fv3_ccpp_regional_control and fv3_ccpp_regional_restart test in develop branch.

tests/tests/fv3_ccpp_gfs_v15p2_coldstart
```diff
@@ -0,0 +1,48 @@
...
+
+export CNTL_DIR=fv3_gfs_v15p2
```
DusanJovic-NOAA, line 9: Is this correct CNTL_DIR? Shouldn't it be fv3_gfs_v15p2_ccpp?

tests/tests/fv3_ccpp_gfs_v15p2_restart
```diff
@@ -0,0 +1,90 @@
+###############################################################################
+#
+#  FV3 CCPP GFS v15.2 compiled with 32-bit dynamics restart (24-48h) test
+#
+###############################################################################
+
+export TEST_DESCR="Compare FV3 32bit CCPP GFS v15.2 results with previous trunk version"
+
+export CNTL_DIR=fv3_gfs_v15p2
```
DusanJovic-NOAA, line 9: Is this correct CNTL_DIR?

tests/tests/fv3_ccpp_gfs_v15p2_coldstart
climbfuji: The CNTL_DIR is correct, the run script appends the RT_SUFFIX environment variable so that this test works for IPD PROD, CCPP PROD, IPD/CCPP REPRO.

tests/tests/fv3_ccpp_gfs_v15p2_restart
climbfuji: ditto

tests/tests/fv3_ccpp_gfs_v15p2_coldstart
DusanJovic-NOAA: I see. Thanks.

build.sh
```diff
@@ -3,8 +3,10 @@ set -eu
 
 MYDIR=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )" && pwd -P)
 
-export COMPILER=${COMPILER:?"Please set COMPILER environment variable [gnu|intel]"}
-export CMAKE_Platform=linux.${COMPILER}
+if [ -z ${CMAKE_Platform+x} ]; then
```
climbfuji: @DusanJovic-NOAA please have a look at this proposed change and let me know if it is ok for you. Using this if test allows for the following on platforms supported by the ufs-weather-model:
```
module use $PWD/modulefiles/hera.intel/fv3
./build.sh
```
One could of course also require the user to set `CMAKE_Platform` directly and remove the `COMPILER` logic.

tests/rt.conf
```diff
@@ -2,20 +2,24 @@
...
+COMPILE | 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                   | standard    | hera.intel     | fv3         |
+COMPILE | 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                   | standard    | cheyenne.intel | fv3         |
+COMPILE | 32BIT=Y CCPP=Y STATIC=Y SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta                                                                   | standard    | cheyenne.gnu   | fv3         |
 
 RUN     | fv3_ccpp_gfs_v15p2                                                                                                             | standard    |                | fv3         |
+RUN     | fv3_ccpp_gfs_v15p2_coldstart                                                                                                   | standard    |                |             |
+RUN     | fv3_ccpp_gfs_v15p2_restart                                                                                                     | standard    |                |             | fv3_ccpp_gfs_v15p2_coldstart
```
climbfuji, line 11: @DusanJovic-NOAA thanks for the note, I think I added the coldstart dependency as required for both v15p2 and v16beta

climbfuji:
This PR is tested together with https://github.com/NOAA-EMC/fv3atm/pull/45, i.e. need to update the fv3atm submodule pointer after https://github.com/NOAA-EMC/fv3atm/pull/45 is merged.

build.sh
DusanJovic-NOAA: Looks correct to me. We should always try to make build systems as simple as possible. I also agree that ``COMPILER`` can be removed, since it looks like it is only used to set  ``CMAKE_Platform``. I have no idea why I set it up this way. I guess I just reused the script I was using for gnu make builds.

DusanJovic-NOAA (approved):

climbfuji:
Thanks, @DusanJovic-NOAA. I am seeing one issue when running `build.sh` on Cheyenne (it's got to do with their "homemade" MPI wrappers), it doesn't show up when using `compile_cmake.sh` or `rt.sh`. Once I figured that out the PR is good to merge.

climbfuji:
> Thanks, @DusanJovic-NOAA. I am seeing one issue when running `build.sh` on Cheyenne (it's got to do with their "homemade" MPI wrappers), it doesn't show up when using `compile_cmake.sh` or `rt.sh`. Once I figured that out the PR is good to merge.

It turns out that this was a problem with the node and/or wrapper script on Cheyenne, not with the code. Please go ahead with the merge. Thanks!

junwang-noaa (approved):

tests/rt.conf
climbfuji: As per our email conversation, yes, the difference is only the forecast length.

tests/tests/fv3_ccpp_gfs_v15p2_coldstart
climbfuji: As long as there is only one set or restart files it is ok. But I agree that for the develop branch we should use Dusan's approach.

PR #41, opened 2020-01-30
Ufs release v1.0

mark-a-potts:
This is an update that will allow the UFS to build with the new NCEPLIBS and NCEPLIBS-external projects. It attempts to keep the build.sh capability intact while removing the need to specify environment variables by hand when building on a non-NOAA machine. The build process is as follows --

to build from scratch (only compiler, mpi and cmake available)

configure NCEPLIBS-external with "cmake -DMPITYPE=mpitype -DBUILD_ALL=ON path-to-NCEPLIBS-external"
make -j 8

configure NCEPLIBS with "cmake -DEXTERNAL_LIBS_DIR=path-to-NCEPLIBS-external/build/install path-to-NCEPLIBS"
make -j 8
make install <---- This is required

configure ufs-weather-model with "cmake -DEXTERNAL_LIBS_DIR=path-to-NCEPLIBS-external/build/install -DNCEPLIBS_INSTALL_DIR=path-to-NCEPLIBS/build/install path-to-ufs-weather-model"
make -j 8


arunchawla-NOAA:
Mark

These are significant changes to the build system so close to creating final tags. Is this PR addressing a significant short coming of the build system ? My feeling is that the build system as it is right now is very solid and has been ported to multiple platforms easily. It would be better to keep this to the side and come back to it after the release for further discussions. Are there some significant drawbacks of the current build that are being addressed by this PR?

mark-a-potts:
I agree that it is a significant change and am fine if it gets delayed.

From my perspective, the downside of the current system is that it does not have any capability to find decencies on its own and the build.sh script is cumbersome to adjust to new  platforms. That said, the build.sh script does work well on the platforms it is set up for and I tried to maintain that functionality.

I will defer to others judgement on when to implement.

Mark

ceceliadid:
Mark, I appreciate the invitation to review but don't feel I should weigh in - I leave it to others with more expertise in the details of the build. Thanks for your efforts - all changes to improve portability will be valuable even if they come in after the release.

DusanJovic-NOAA:
No system will have capability to find decencies on its own if dependencies are installed in arbitrary location. Period.
So some assumptions must be made. For example in this PR assumptions are that all third-party libraries are installed under ``EXTERNAL_LIBS_DIR``, all ncep libraries are installed under ``NCEPLIBS_INSTALL_DIR`` and that all library names are what's hard-coded in ``setvars`` file, which is less flexible then the current build.sh which allows location and name of each individual library to be specified separately.
What exactly is cumbersome to adjust to new platforms in build.sh? I do not see any platform specific thing in build.sh. Everything can be configured outside build.sh, at least that's the idea. If there's anything that we missed please let us know.

climbfuji:
I agree with Dusan, we shouldn't try to make the ufs-weather-model and other parts smart about this. We only need to set a handful of variables, and we should do this as consistent as possible with what the current method is. This will allow users to swap in and out different versions of a specific NCEPLIBS library, install libraries as he/she wants, it works on preconfigured platforms, ... just my 2 cents.

PR #47, opened 2020-02-06
updated NEMS to fix openmpi crash when MPI_Comm_Split hash is negative

mark-a-potts:
Small change to make the hash used in MPI_Split_Comm call an absolute value and always be positive (openmpi fails with negative number)


DusanJovic-NOAA:
Why is FV3 submodule updated in this PR?

mark-a-potts:
Working too fast. This should be the correct update.

PR #49, opened 2020-02-06
fix openmpi

MinsukJi-NOAA:
Mark's change for openmpi in src/module_NEMS_Rusage.F90.


MinsukJi-NOAA:
Closing this because it is the same as PR #48.

PR #54, opened 2020-02-10
Bugfixes for macOS build

climbfuji:
- turn off OpenMP if the clang C compiler is used (broken since macOS Mojave)
- fix rpath issue for netCDF libraries (there may be a follow-up PR with a better solution - time permitting)


climbfuji:
@kgerheiser @jedwards4b FYI

DusanJovic-NOAA (approved):

PR #60, opened 2020-02-18
Used reset to compute full Thompson refl only on output times, pass k…

ericaligo-NOAA:
…dt into Thompson and allow Thompson run without aerosols.


junwang-noaa:
This PR is committed in PR#84. Close PR.

PR #63, opened 2020-02-21
Consistent use of logicals in namelists

DusanJovic-NOAA:
- Rename .F. to .false. and .T. to .true. in all input.nml templates
- Rename configure_jet.cmake -> configure_jet.intel.cmake
- Remove (unused) 'amip_interp_nml' namelist from all input.nml templates
- Convert files in parm directory from "UTF-8 Unicode (with BOM)" to ASCII text


junwang-noaa (approved):

climbfuji (approved):
This looks good to me! I think we can go ahead and merge it without running the Cheyenne test pre-commit. I'll run them post commit if that's ok.

PR #64, opened 2020-02-21
Sm jan102020

SMoorthi-emc:
This is pull request for the ufs-weather-model app associated with fv3atm pull request for 
https://github.com/SMoorthi-emc/fv3atm branch SM_Jan102020 which includes updates to two way physics for two way wave-atmosphere coupling (both in ccpp and ipd).  It also includes a fix in gcycle for update to lake temperature climatology, and a fix to the mask exported to the NEMS Mediator for the interpolation of atmospheric fields over ocean points only. Associated with this PR is a PR updating the ccpp physics.


junwang-noaa:
Code is committed, close PR.

PR #128, opened 2020-05-20
initialize ice fluxes and add "tiice" array. stochastic_physics

DusanJovic-NOAA:
Associated PRs:

FV3: https://github.com/NOAA-EMC/fv3atm/pull/119
stochastic_physics: https://github.com/noaa-psd/stochastic_physics/pull/23


climbfuji (changes requested):
I am fine with the stochastic_physics PR, but the fv3atm PR - aside from the question whether we accept changes to IPD or not - has some issues: (1) the newly introduced ice temperature doesn't get used anywhere in the physics, it is just added and read in/written out (2) there is no submodule pointer update for the ccpp-physics PR. @shansun6

climbfuji (approved):

junwang-noaa (approved):

PR #133, opened 2020-05-26
Have CMake ensure that the fortran compiler supports the intrinsic FINDLOC

### LINKED ISSUE #132: Have CMake check capability for handling Fortran 2008 intrinsic FINDLOC
See discussion in #114.

Seems that we need the GNU fortran compiler to handle some aspects of 2003 which it does not do before version 9.0.

So the CMake build needs to check the fortran compiler to make sure it can handle the Fortran 2003 that it needs to handle.

edwardhartnett:
The FINDLOC intrinsic is used in the code, but is not supported by all versions of all fortran compilers.

In this PR I add a check for FINDLOC in the CMake build. If it can't be found, the build exits with an error message.

Fixes #132

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ add_definitions(-DINTERNAL_FILE_NML)
 
 include(CheckFortranSourceCompiles)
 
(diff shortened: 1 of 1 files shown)


edwardhartnett:
It has been decided to solve this is a different way. @DusanJovic-NOAA would you like to prepare the PR or should I?

PR #134, opened 2020-05-27
checking version of gfortran compiler

### LINKED ISSUE #132: Have CMake check capability for handling Fortran 2008 intrinsic FINDLOC

edwardhartnett:
Fixes #132 

This is the GNU-specific solution.


climbfuji:
This PR has been pulled into and will be merged with https://github.com/ufs-community/ufs-weather-model/pull/139.

PR #147, opened 2020-06-15
Update modulefile for hera.gnu

climbfuji:
The GNU software stack on Hera doesn't work with the ufs-weather-model. When compiled with the openmpi/3.1.4 module the model hangs indefinitely after initialization.

This PR updates the hera.gnu modulefile to use an mpich-3.3.2 build (and subsequent netCDF, ESMF and NCEPLIBS builds) that I created under the shared BMC/gmtb space.

The model runs are relativelty slow, presumably because I built mpich-3.3.2 without any tailoring to the Hera Infiniband fabric, but at least it runs. A C96 regression test finishes in about 10 minutes.

Note that I had to test the updated modulefile with an earlier version of the ufs-weather-model code, because the current code does not compile with GNU (FV3 dycore updates).

### DIFF
--- a/modulefiles/hera.gnu/fv3
+++ b/modulefiles/hera.gnu/fv3
@@ -23,13 +23,12 @@ module load sutils
 ## this typically includes compiler, MPI and job scheduler
 ##
 module load gnu/9.2.0
-module load openmpi/3.1.4
-module load netcdf/4.7.2
+module use -a /scratch1/BMC/gmtb/software/modulefiles/gnu-9.2.0/mpich-3.3.2
+module load mpich/3.3.2
 
 ##
-## use pre-compiled EMSF library and NCEP libraries for above compiler / MPI combination
+## use pre-compiled netCDF, EMSF library and NCEP libraries for above compiler / MPI combination
 ##
-module use -a /scratch1/BMC/gmtb/software/modulefiles/gnu-9.2.0/openmpi-3.1.4
 module load NCEPlibs/1.0.0
 
 ##


MinsukJi-NOAA:
I wonder if mvapich 2/2.3 could make it faster -- it says it is compatible with mpich 3.2 (https://mvapich.cse.ohio-state.edu/downloads/)

climbfuji:
> I wonder if mvapich 2/2.3 could make it faster -- it says it is compatible with mpich 3.2 (https://mvapich.cse.ohio-state.edu/downloads/)

From my experience, the differences are marginal. It depends mostly on how you configure the MPI library. I also do not intend to maintain that MPI library build on the long term, this is something the sysadmins should do. Compiling an MPI library with the correct settings for performance requires a little more than compiling a netCDF library. You need to know some of the details of the underlying fabric. It's already tricky to build it with the slurm integration (which I did).

DusanJovic-NOAA (approved):

junwang-noaa (approved):
I assume the fix was tested with the latest model.

climbfuji:
> I assume the fix was tested with the latest model.

The change is only for a gnu modulefile that isn't used by the code when compiling on another machine or with Intel on hera, regression tests are not impacted at all. As described above in the description of the PR (https://github.com/ufs-community/ufs-weather-model/pull/147#issue-4348208650) I had to copy the updated modulefile from this commit into a previous version of the code (before FV3 dycore commit) in order to test compiling/running the regression tests on hera.gnu (because the FV3 dycore commit breaks the GNU build).

climbfuji:
This PR was merged as part of https://github.com/ufs-community/ufs-weather-model/pull/151

Issue #142, opened 2020-06-04
update esmf lib to esmf 8.1.0bs20


junwang-noaa:
ESMF 8.1.0bs20 is available with features required for coupled model and the performance improvement in esmf 8.0.1 used in GFS.v16. ufs-weather-model will be updated to use the latest esmf 8.1.0bs20. The code changes to improve GFS.v16 performance will be added in fv3 and nems.

PR #166, opened 2020-07-14
Ice flux initialization in the cold run

ShanSunNOAA:
When ice flux is huge (undefined in CICE yet) in the 1st time cold run, use PBL-calculated ice fluxes instead. This will change results in the coupled model only. Passed regression tests for standalone FV3.

This refers to FV3 issue #142 and CCPP issue #472.

@junwang-noaa


junwang-noaa:
The associated FV3 and ccpp PR is closed, so close this PR.

Issue #168, opened 2020-07-16
implement github actions workflow for carrying out utests


MinsukJi-NOAA:
* as a part of continuous integration framework, use github actions to carry out utests upon pull request to the weather-model repository

PR #169, opened 2020-07-16
github actions utest

MinsukJi-NOAA:
This PR addresses issue #168 
* Include continuous integration related files (scripts, github actions yaml file, dockerfile, etc.)
* Include updates to the utest script


.github/workflows/main.yml
```diff
@@ -0,0 +1,145 @@
...
+      tn: ${{ steps.parse.outputs.tn }}
+      bld: ${{ steps.parse.outputs.bld }}
+      test: ${{ steps.parse.outputs.test }}
+      img: ${{ steps.parse.outputs.img }}
+
+    steps:
+    - name: Checkout codes
+      uses: actions/checkout@v2
+
+    - name: Parse cases
```
DusanJovic-NOAA, line 26: Why is all this parsing of ci.test file done in action instead of in some script that action just runs. The action file should not contain any logic about how tests are done and what they are actual doing. It's (action) purpose should be just to run some script based on some repository event.

.github/workflows/main.yml
```diff
@@ -0,0 +1,145 @@
...
+        case_=$(sed -n 2p ${GITHUB_WORKSPACE}/tests/ci/ci.test)
+        img_=$(sed -n 3p ${GITHUB_WORKSPACE}/tests/ci/ci.test)
+
+        imd1_=''
+        [[ $case_ =~ thr || $case_ =~ mpi || $case_ =~ dcp || $case_ =~ rst ]] && imd1_+='std'
+        [[ $case_ =~ bit ]] && imd1_+=' bit'
+        [[ $case_ =~ dbg ]] && imd1_+=' dbg'
+        imd1_=$(echo $imd1_ | sed -e 's/^ *//' -e 's/ *$//')
+
+        bld_='{"bld_set":['
```
DusanJovic-NOAA: This looks like you are constructing a json object, in a shell script based on option parsed from ct.test. Is this what's going on here?

.dockerignore
```diff
@@ -0,0 +1,5 @@
+Dockerfile
+.git
+.gitignore
+tests/*.log
```
DusanJovic-NOAA, line 4: Do not ignore all .log files from tests directory.

DusanJovic-NOAA: Sorry, I thought this is a .gitignore.

.github/workflows/main.yml
MinsukJi-NOAA: Yes, that is exactly right.
- In ci.test, a user specifies which unit tests to run: thr, mpi, dcp, rst, dbg, bit
- Based on this, json objects are constructed which includes information on which cases to build and run
- The json objects are used in the subsequent steps below to run actions in a matrix

MinsukJi-NOAA:
RT on Orion keeps failing, due to compile_13 failing with this error message:

`/work/noaa/stmp/jminsuk/stmp/jminsuk/FV3_RT/rt_420215/compile_13/build_fv3_13/FV3/ccpp/physics/ccpp_static_api.F90(966): catastrophic error: **Internal compiler error: internal abort** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
compilation aborted for /work/noaa/stmp/jminsuk/stmp/jminsuk/FV3_RT/rt_420215/compile_13/build_fv3_13/FV3/ccpp/physics/ccpp_static_api.F90 (code `1)`

climbfuji:
> RT on Orion keeps failing, due to compile_13 failing with this error message:
> 
> `/work/noaa/stmp/jminsuk/stmp/jminsuk/FV3_RT/rt_420215/compile_13/build_fv3_13/FV3/ccpp/physics/ccpp_static_api.F90(966): catastrophic error: **Internal compiler error: internal abort** Please report this error along with the circumstances in which it occurred in a Software Problem Report. Note: File and line given may not be explicit cause of this error. compilation aborted for /work/noaa/stmp/jminsuk/stmp/jminsuk/FV3_RT/rt_420215/compile_13/build_fv3_13/FV3/ccpp/physics/ccpp_static_api.F90 (code `1)`

Hmm. I have seen these internal compiler errors for other projects in the past, it often had to do with a too complicated optimization task that the compiler is trying to accomplish. We need to figure out which COMPILE command build_13 corresponds to, and what line the compiler is complaining about. Which version of the Intel compiler are you using?

MinsukJi-NOAA:
> Hmm. I have seen these internal compiler errors for other projects in the past, it often had to do with a too complicated optimization task that the compiler is trying to accomplish. We need to figure out which COMPILE command build_13 corresponds to, and what line the compiler is complaining about. Which version of the Intel compiler are you using?

modulefiles/orion.intel/fv3 module loads intel/2018

compile 13 appears to be CCPP=Y and DEBUG=Y

climbfuji:
> > Hmm. I have seen these internal compiler errors for other projects in the past, it often had to do with a too complicated optimization task that the compiler is trying to accomplish. We need to figure out which COMPILE command build_13 corresponds to, and what line the compiler is complaining about. Which version of the Intel compiler are you using?
> 
> modulefiles/orion.intel/fv3 module loads intel/2018
> 
> compile 13 appears to be CCPP=Y and DEBUG=Y

Interesting. I recently made this change (and it worked on all machines, including orion) so that we have at least one compile command that tests compiling the code without providing any suites (i.e. compile all available suites). What you can do is to duplicate this line N times (where N is the number of machines, i.e. one line for hera.intel, one for cheyenne.intel, ...) and modify the line for orion to include just the suites that are required to run the following tests.

MinsukJi-NOAA:
> Interesting. I recently made this change (and it worked on all machines, including orion) so that we have at least one compile command that tests compiling the code without providing any suites (i.e. compile all available suites). What you can do is to duplicate this line N times (where N is the number of machines, i.e. one line for hera.intel, one for cheyenne.intel, ...) and modify the line for orion to include just the suites that are required to run the following tests.

That change allowed the RT to pass. Thanks.

MinsukJi-NOAA:
Preliminary CI documentation can be found here: [CI tests for UFS-weather-model](https://docs.google.com/presentation/d/1nNTbLkLYWGKYnHwfLk99zW_MSztiURvv3INn6MRNfHY/edit?usp=sharing)

MinsukJi-NOAA:
For an example of a CI test, see [Add hera RT results](https://github.com/MinsukJi-NOAA/ufs-weather-model/actions/runs/233381117)

For an example of a skipped CI test, see [Add orion RT results](https://github.com/MinsukJi-NOAA/ufs-weather-model/actions/runs/234268549)

modulefiles/linux.gnu/fv3
```diff
@@ -24,3 +37,19 @@ if [ ! "x$SIONLIB" == "x" ]; then
...
+
+##
+## use pre-compiled EMSF library for above compiler / MPI combination
+##
+export ESMFMKFILE=${ESMFMKFILE:-/usr/local/lib/esmf.mk}
+
+##
+## NCEP libraries (need to download and build manually, see doc/README_{UBUNTU,CENTOS,...}.txt and https://github.com/NCAR/NCEPlibs)
+##
+export NCEPLIBS_DIR=${NCEPLIBS_DIR:-/usr/local/NCEPlibs}
```
climbfuji, line 49: This looks like the "old" (UFS MRW 1.0/1.1) way to use the libraries. I assume you have some more or less recent hash of the "new" NCEPLIBS (i.e. develop branch) that could be used as imported targets, right? Just curious, no need to change anything.

.github/workflows/manage.yml
```diff
@@ -0,0 +1,54 @@
+name: Manage workflows
```
DusanJovic-NOAA, line 1: Why is this necessary?

.github/workflows/main.yml
climbfuji: I agree, most of this could be encapsulated in some script (in whatever preferred scripting language) to simplify these `.yml` files.

.github/workflows/main.yml
DusanJovic-NOAA: Too complicated. Like I said in my previous comment I do not think github action should contain any logic that depends on specification of ci.test file.

tests/ci/json_helper.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
```
climbfuji, line 1: I am not familiar with json, unfortunately, hence my probably dumb question: since you are wrapping some of the json logic in Python, wouldn't it be possible to do everything in Python?

tests/run_test.sh
```diff
@@ -130,11 +130,20 @@ atparse < ${PATHTR}/parm/${NEMS_CONFIGURE:-nems.configure} > nems.configure
 # Submit test job
 ################################################################################
 
-if [[ $ROCOTO = 'false' ]]; then
-  submit_and_wait job_card
+if [[ $SCHEDULER = 'none' ]]; then
```
climbfuji, line 133: I believe you just facilitated my next job (run official regression tests in the cloud) by quite a bit!

.github/workflows/manage.yml
MinsukJi-NOAA: Our current PR process involves a developer committing multiple times after a PR is open. We do not want each of them to start up a new GitHub Actions. The functions of `manage.yml` are twofold: 1) when GitHub Actions is started, if it sees previously running (or queued) tests of the same same user and head branch, it will cancel them; 2) if a developer specify skip-ci in his/her commit message, GitHub Actions will not be started.

Because of the write-permission issues in GitHub Actions, these functions cannot be included in `main.yml`, and thus needs a second yml file.

modulefiles/linux.gnu/fv3
MinsukJi-NOAA: I kept up to date with the latest development in the weather model. However, I think one important thing is for us to maintain a linux.gnu set of tools and libraries; I am thinking that they will need to be version controlled and maintained for any cloud (and docker) related runs.

tests/run_test.sh
MinsukJi-NOAA: Cool. You will need lots of resources to run RT in the cloud!

.github/workflows/main.yml
aerorahul: I agree with @DusanJovic-NOAA. There is too much `scripting` in these actions.

MinsukJi-NOAA: OK. Let me try putting the parsing section in a separate script.

tests/ci/json_helper.py
MinsukJi-NOAA: `manage.yml` gets input from GitHub API; then `json_helper.py` goes through the json logic and processes the input, and produces output; then `manage.yml` uses the output to interact with GitHub API. As GitHub API supports multiple languages, may it can be all done in Python, or by an addition of Java script or something like that; unfortunately, I am not an expert either, and will take some effort to implement and test.  I believe `manage.yml`, as it is, is not too complex.

climbfuji: Thanks for the explanation. I don't know and understand json, and it seems complicated to use multiple nested languages. But if you are comfortable and familiar with it, then that's fine.

MinsukJi-NOAA: Bash has `jq` tool to handle json. But it needs to be installed, so Python 3 was chosen instead. Maybe in the future, Python 3 will handle GitHub API, thus simplifying `manage.yml`

DusanJovic-NOAA: The main question here is do we need to do all of this. Can't we just have a simple action file that runs a container and in that container we run a script that in turn runs ci.sh which executes whatever needs to be executed. Simple as that. No messing with GitHub API, no parsing ci.test in yml, no creation of json, no parsing of that json in python, no manage.yml, nothing. Just a shell script that runs in a container that returns success of fail exit code. That's all.

MinsukJi-NOAA: I recall @DusanJovic-NOAA did not want GitHub Actions to be fired up every time a commit is made during our typical PR process, where multiple commits are common before a PR is merged. So, I addressed that issue with GitHub API and extra yml file and python script.

DusanJovic-NOAA: Exactly. That's why I think these kind of tests (tests that run longer than few minutes) are simply not suitable for these "continuous integration" testing environments.

junwang-noaa:
@DusanJovic-NOAA I remember Dom said the github action running time for
public repo is unlimited, also because of the characteristics of weather
model forecast tests, we have to run jobs longer than a few minutes, so
what are the issues?

climbfuji:
> @DusanJovic-NOAA I remember Dom said the github action running time for public repo is unlimited, also because of the characteristics of weather model forecast tests, we have to run jobs longer than a few minutes, so what are the issues?
> […](#)

Have a look at https://github.com/pricing, scroll down to "Compare features" please.

MinsukJi-NOAA:
Updated utest documentation can be found here: [Unit Test for UFS-weather-model](https://docs.google.com/presentation/d/1fmTrF-_s7oseswlfGOlSkMTdRbG6yJ59ExZ7kZvJuxk/edit#slide=id.p)

.github/workflows/main.yml
MinsukJi-NOAA: The parsing section has been moved to a separate script to simplify.

MinsukJi-NOAA:
All regression tests passed on Hera, Orion, and Dell. All unit tests except restart test passed on Hera, Orion, and Dell. This PR is ready for merge.

climbfuji (approved):
Looks ok to me and my limited understanding of the CI design.

tests/utest
```diff
@@ -182,116 +270,145 @@ run_utests() {
 					export LIST_FILES="${LIST_FILES}"
 					EOF
         ;;
-      32bit)
-        comp_nm=$rc
+      bit_base)
```
junwang-noaa, line 273: Is the bit_base run to create 32bit baseline and then the bit test runs again the baseline?

junwang-noaa (approved):

tests/utest
MinsukJi-NOAA: By default, Bit and Debug only check if code compiles and runs to completion. There is, however, an option to check the reproducibility of Bit (-b) and Debug (-d).

DusanJovic-NOAA (approved):

PR #172, opened 2020-07-21
Update CMakeLists.txt files to use imported targets

DusanJovic-NOAA:
## Description

Top level CMakeLists.txt and CMakeLists.txt file in FV3 have been changed (and cleaned up) to use imported targets (mostly for NCEPLIBS).
Remove *.appBuilder compsets and log.
Remove outdated doc directory

## Testing

Full Regression tests have been run on WCOSS (Cray and Dell), Hera (GNU and Intel) and Orion.
GNU tests on Hera require new baselines, as well as test on WCOSS Cray that runs inline post.

## Dependencies

- waiting on noaa-emc/fv3atm/pull/146


CMakeLists.txt
```diff
@@ -18,11 +18,51 @@ project(ufs-weather-model
...
+
+find_package(bacio REQUIRED)
+find_package(nemsio REQUIRED)
+find_package(sp REQUIRED)
+find_package(w3emc REQUIRED)
+find_package(w3nco REQUIRED)
+if(INLINE_POST)
+  find_package(nceppost REQUIRED)
+endif()
+# See https://github.com/NOAA-EMC/NCEPLIBS-nemsio/pull/22
```
aerorahul, line 79: This was merged. Is a comment to a PR still necessary?  
May be more appropriate to say, why mixed types are being used.

DusanJovic-NOAA: Yes. That PR was merged. I added that comment here just to remind us why we need to explicitly define nemsio's dependency on specific versions of w3emc and bacio.

CMakeLists.txt
```diff
@@ -210,34 +163,24 @@ if(WW3)
...
   message("Build WW3:")
-  message("   run: ${CMAKE_BUILD_TOOL} WW3_PARCOMPN=4 WW3_COMP=${WW3_COMP} ww3_nems")
+  message("   run: ${CMAKE_BUILD_TOOL} WW3_PARCOMPN=4 WW3_COMP=${WW3_COMP} ww3_nemslibonly")
   message("   in:  ${PROJECT_SOURCE_DIR}/WW3/model/esmf")
   message("")
 
   add_custom_target(ww3_nems
     WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/WW3/model/esmf
-    COMMAND ${CMAKE_BUILD_TOOL} WW3_PARCOMPN=4 WW3_COMP=${WW3_COMP} ww3_nems > make.log 2>&1
+    COMMAND ${CMAKE_BUILD_TOOL} WW3_PARCOMPN=4 WW3_COMP=${WW3_COMP} ww3_nemslibonly > ${PROJECT_BINARY_DIR}/ww3_make.log 2>&1
```
junwang-noaa, line 188: Does this code use ww3 cmake build that you recently committed?

climbfuji (approved):
Works for me, also on my macOS with NCEPLIBS-external and NCEPLIBS. Great work!

CMakeLists.txt
aerorahul: @junwang-noaa  No.

CMakeLists.txt
```diff
@@ -250,41 +193,79 @@ add_executable(
...
+
+###############################################################################
+### UFS executable
+###############################################################################
+add_executable(NEMS.exe
+               NEMS/src/MAIN_NEMS.F90
+)
+add_dependencies(NEMS.exe ufs-weather-model-lib)
+
+target_compile_definitions(NEMS.exe PRIVATE -DESMF_VERSION_MAJOR=${ESMF_VERSION_MAJOR})
```
aerorahul, line 256: Is this needed? Wouldn't it inherit from `ufs-weather-model-lib`?

CMakeLists.txt
```diff
@@ -210,34 +163,24 @@ if(WW3)
...
-configure_file(${PROJECT_SOURCE_DIR}/NEMS/src/ESMFVersionDefine_ESMF_NUOPC.h
-               ${PROJECT_BINARY_DIR}/NEMS/ESMFVersionDefine.h
-               COPYONLY)
-configure_file(${PROJECT_SOURCE_DIR}/NEMS/src/ESMFConvenienceMacros.h
-               ${PROJECT_BINARY_DIR}/NEMS/src/ESMFConvenienceMacros.h
-               COPYONLY)
-
-add_executable(
-    NEMS.exe
+add_library(
```
aerorahul, line 197: Can this be made into `libnems` consisting solely of the NEMS source code?

`ufs-weather-model-lib` is a collection of `libnems, fv3atm, ww3_nems`
It will simplify the logic under hear as everything is `target_link_libraries` or `target_compile_definitions` to `ufs-weather-model-lib`.

CMakeLists.txt
DusanJovic-NOAA: It is needed at this moment, because `ESMF_VERSION_MAJOR` is added as `PRIVATE` definition to `ufs-weather-model-lib`. We can add it as `PUBLIC` but then all users of ufs-weather-model-lib (ie. JEDI) will get it. Do we want that? I thought it makes more sense to keep it private. In any case this will all be replaced pretty soon.

cmake/configure_hera.gnu.cmake
```diff
@@ -1,37 +1,11 @@
...
-message("Fortran compiler: ${CMAKE_Fortran_COMPILER_ID} ${CMAKE_Fortran_COMPILER_VERSION} (${Fortran_COMPILER_NAME})")
-message("")
-
-option(DEBUG   "Enable DEBUG mode" OFF)
-option(REPRO   "Enable REPRO mode" OFF)
-option(VERBOSE "Enable VERBOSE mode" OFF)
-option(32BIT   "Enable 32BIT (single precision arithmetic in dycore)" OFF)
-option(OPENMP  "Enable OpenMP threading" ON)
-option(AVX2    "Enable AVX2 instruction set" OFF)
+option(32BIT           "Enable 32BIT (single precision arithmetic in dycore)" OFF)
```
aerorahul: rather than have the same information in all the `configure_${PLATFORM}.${compiler}.cmake`,
can we reduce this to 2 sets?
`configure.gnu.cmake` and `configure.intel.cmake`.
If there is a platform that has something that needs something different that the "stock" options, only those options can be placed in a file with `${PLATFORM}`

CMakeLists.txt
DusanJovic-NOAA: Are you suggesting to rename  `ufs-weather-model-lib` to `nems`? I can do that. But I'm not sure I know how to make `ufs-weather-model-lib` a collection of `nems`, `fv3atm`, and optionally `ww3_nems`.

aerorahul: yes.
you can create ufs-weather-model-lib from nems, fv3atm etc. same way you created fv3atm from fv3cpl, gfsphys, etc.

cmake/configure_hera.gnu.cmake
DusanJovic-NOAA: We (@climbfuji and @junwang-noaa) discussed this and decided that to keep all options defined in separate files for all 'Tier-1' platforms. Your suggestion tho have `configure.gnu.cmake` and `configure.intel.cmake` is basically what is now `configure_linux.gnu.cmake` and `configure_linux.intel.cmake` (ie. generic Linux platforms)

aerorahul: and why that is not sufficient? Are Tier-1 platforms not built on generic Linux platforms, give or take some options e.g. AVX2 or inline POST

CMakeLists.txt
DusanJovic-NOAA: But `fv3atm`, in addition to depending on `fv3cpl`, `gfsphysics` also contains code (`atmos_model.F90`, `fv3_cap.F90`, ...).  How do I create library without code and then add other libraries as dependencies. I guess I do not understand what you mean by 'collection'? I do not know how to "_create ufs-weather-model-lib from nems, fv3atm_". I can only add `nems` and `fv3atm` as dependencies to `ufs-weather-model-lib`.

cmake/configure_hera.gnu.cmake
climbfuji: When we discussed this initially, the suggestion was to only have the tier-1 platform configuration and users would have to always specify deviations from it on the command line. This is not ok, the model should build out of the box on other systems as well. Rahul's suggestion will work, and so does Dusan's. As long as users do not have to manually specify `-DINLINE_POST=OFF` every time the build the model, I don't care which version we use.

DusanJovic-NOAA: My original suggestion was to just put all options in the top-level `CMakeLists.txt`. Because `configure_linux.gnu.cmake` and `configure_linux.intel.cmake` are identical we do not need them both.

aerorahul: @DusanJovic-NOAA 
I second that suggestion.
Looking ahead with a coupled model, these options are just going to increase and become too many. And placing/repeating them in all platform/tier specific files is going to get messy, hard to maintain and prone to errors.

DusanJovic-NOAA: How about this. In the top-level `CMakeLists.txt` we define these options (with the default values equal to what's currently in 'generic' linux configuration files):
```
set(32BIT           OFF CACHE BOOL "Enable 32BIT (single precision arithmetic in dycore)")
set(AVX2            ON  CACHE BOOL "Enable AVX2 instruction set")
set(CCPP            ON  CACHE BOOL "Enable CCPP")
set(DEBUG           OFF CACHE BOOL "Enable DEBUG mode")
set(INLINE_POST     OFF CACHE BOOL "Enable inline post")
set(MULTI_GASES     OFF CACHE BOOL "Enable MULTI_GASES")
set(OPENMP          ON  CACHE BOOL "Enable OpenMP threading")
set(PARALLEL_NETCDF OFF CACHE BOOL "Enable parallel NetCDF")
set(QUAD_PRECISION  ON  CACHE BOOL "Enable QUAD_PRECISION (for certain grid metric terms in dycore)")
set(REPRO           OFF CACHE BOOL "Enable REPRO mode")
set(WW3             OFF CACHE BOOL "Enable WW3")
```
Then based on `CMAKE_Platform`, if file exists, we include platform specific file:
```
if(EXISTS ${PROJECT_SOURCE_DIR}/cmake/configure_${CMAKE_Platform}.cmake)
  message("")
  message("Setting configuration for $ENV{CMAKE_Platform}")
  message("")
  include(${PROJECT_SOURCE_DIR}/cmake/configure_${CMAKE_Platform}.cmake)
endif()
```
On Hera, for example, this file will be:
```
$ cat cmake/configure_hera.intel.cmake 
set(INLINE_POST     ON CACHE BOOL "Enable inline post" FORCE)
set(PARALLEL_NETCDF ON CACHE BOOL "Enable parallel NetCDF" FORCE)
```
which will override `INLINE_POST` and `PARALLEL_NETCDF`.

climbfuji: That's fine with me. How about compiler-dependent options? AVX2 only makes sense if Intel is used, for GNU that option doesn't exist (or could be replaced by another option).

DusanJovic-NOAA: We do not use AVX2 option to define architecture specific compiler flags for GNU, but GNU compilers do have such options (`-march=...`). Maybe we'll start using them.
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

modulefiles/wcoss_cray/fv3
```diff
@@ -31,23 +32,27 @@ module load gcc/4.9.2
...
+module load ip/3.3.0
+module load nceppost/dceca26
+module load nemsio/2.5.1
+module load sp/2.3.0
+module load w3emc/2.7.0
+module load w3nco/2.4.0
+
+module load gfsio/1.4.0
+module load sfcio/1.4.0
+module load sigio/2.3.0
```
junwang-noaa, line 54: May I ask not to load gfsio, sfcio and sigio? Model does not need those extra old libraries? Also what is nceppost/dceca26, why do we need to add them to model module file?

junwang-noaa (approved):
I am not sure why gfsio/sfcio/sigio are added to model module files, if these are not required, consider to remove them in next commit. Other than this, the changes look good to me.

cmake/FindNetCDF.cmake
```diff
@@ -0,0 +1,337 @@
...
+
+    set( ${_prefix}_${_comp}_LIBRARIES     ${NetCDF_${_comp}_LIBRARIES} )
+    set( ${_prefix}_${_COMP}_LIBRARIES     ${NetCDF_${_comp}_LIBRARIES} )
+    set( ${_prefix}_${_arg_comp}_LIBRARIES ${NetCDF_${_comp}_LIBRARIES} )
+
+    set( ${_prefix}_${_comp}_INCLUDE_DIRS     ${NetCDF_${_comp}_INCLUDE_DIRS} )
+    set( ${_prefix}_${_COMP}_INCLUDE_DIRS     ${NetCDF_${_comp}_INCLUDE_DIRS} )
+    set( ${_prefix}_${_arg_comp}_INCLUDE_DIRS ${NetCDF_${_comp}_INCLUDE_DIRS} )
+  endforeach()
+endforeach()
```
climbfuji, line 337: Looks like this is the same `FindNetCDF.cmake` that is in the CMakeModules repo. Wondering if we should remove the contents of `cmake` in ufs-weather-model in future commit and use https://github.com/NOAA-EMC/CMakeModules as a submodule instead?

cmake/configure_linux.intel.cmake
```diff
@@ -1,37 +0,0 @@
-message("")
```
climbfuji, line 1: I don't see any cleanup of `cmake/configure_macosx.*.cmake`, and several others are also unchanged - is this intentional?

DusanJovic-NOAA: Yes. I do not have macosx to test, so I left it as it is. It should be cleaned up by somebody who can run a test on mac. Also other config files (cheyenne, gaea, stampede) should be updated.

climbfuji: So should we merge it and then commit a cleanup PR for the other platforms within a day, while Moorthi is working on his PR? These changes will be totally independent, won't affect Moorthi's work and any of the tier-1 platforms.

cmake/FindNetCDF.cmake
DusanJovic-NOAA: We can not remove cmake directory, as it contains config files and Intel.cmake and GNU.cmake with compiler flags. Those are ufs-weather-model specific.

climbfuji: > We can not remove cmake directory, as it contains config files and Intel.cmake and GNU.cmake with compiler flags. Those are ufs-weather-model specific.

That's right, of course. Let's think about this later. Every file we duplicate between repositories is bad.

cmake/configure_linux.intel.cmake
aerorahul: What is the long-term strategy for keeping unsupported platforms?
This should not be a 2 step process where supported platforms are updated in one PR and then the unsupported in another.  This is bound to be unmaintainable.

We either remove the unsupported ones from this repo and make that a responsibility of the org that is supporting these platforms.  PR's should not be held up for platforms that developers have no access to.  NOAA Tier-1 platforms is a must.

cmake/FindNetCDF.cmake
aerorahul: @climbfuji Really? ufs-s2s-model duplicates an entire repository.

cmake/configure_linux.intel.cmake
climbfuji: These platforms are not unsupported, they are tier-2 and tier-3. I am absolutely against removing the configurations that we use for daily work (How many real-time parallels are running on jet?! How much development for CMEPS etc is done on Cheyenne?! DTC relies on generic macOS and Linux platforms to work to do their work, and the community will too). The tier-1,2,3 concept makes sense as defined - tier-1 testing always pre-commit, tier-2/3 ideally too, but if need be retrospectively. I'd rather hold back this PR for another few hours until I tested the remaining platforms.

cmake/FindNetCDF.cmake
climbfuji: Yes, and as you might be aware of there is an effort to merge it with ufs-weather-model in the future.

cmake/configure_linux.intel.cmake
DusanJovic-NOAA: Yes. These platforms are supported. They are just not in "Tier-1" category, meaning we do not require full regression test to pass before each commit. But we try our best to make sure model is running on those platforms, by providing libraries, etc. That support is not done by EMC, because we do not have access to those machines.

aerorahul: Ok. I understand the concept of tier 1-2-3.
What is the agreed wait period for tier 2-3 platform to be tested. 1 week? Surely that is enough. Who is responsible for that?
What if there is an issue? Who will resolve the issues and how?
Is this information documented somewhere? 
Seems like we are locked in this type of conversation often.

I am not suggesting removing generic configs.  Those are necessary for development on generic platforms.

modulefiles/wcoss_cray/fv3
DusanJovic-NOAA: nceppost/dceca26 is post library.

cmake/configure_linux.intel.cmake
climbfuji: Have a look here https://github.com/ufs-community/ufs-weather-model/wiki/Regression-Test-Policy-for-Weather-Model-Platforms-and-Compilers and here https://github.com/ufs-community/ufs/wiki/Regression-Test-Policy-for-UFS-Platforms-and-Compilers (the second is referenced in the first). Obviously still work in progress to some extend.

PR #179, opened 2020-08-04
Downgrade FMS to 2019.01.03

DusanJovic-NOAA:
## Description

Downgrade FMS to 2019.01.03.
Small changes in cmake build system due to FMS version change.
Improve error handling in rt.sh when ecflow is used. Scripts cleanup and refactoring.

### Issue(s) addressed

In order to merge (unify) ufs-weather-model and ufs-s2s-model changing the FMS to the exact version used by s2s is required.

## Testing

Regression test will be performed on all Tier-1 platforms.
New baseline (develop-20200805) has been created, because of file format change in FMS (CDF2 to HDF5). All restart files and all history files (created using write component) are bit-identical.

## Dependencies

N/A


junwang-noaa (approved):
The code changes look good to me. I noticed that the atmos_4xdaily files are still there, so I assume we will have another commit to clean up that.

climbfuji (commented):
Looks good to me - I'll try this code on cheyenne while we are waiting for hera to come back.

climbfuji (approved):
I tested killing jobs on Cheyenne and it seems to work:
- `kill -15` on the rt.sh job kills all jobs in the queue
- hitting `Ctrl + C` from the rt.sh job itself doesn't kill the queued jobs (is this intended?)

climbfuji:
To complete the testing, I ran the `rt_gnu.conf` tests on Cheyenne against the old/existing baseline. All tests ran to completion within the walltime used usually for cheyenne.gnu but "failed" as expected, due to differences in `atmos_4xdaily*.nc`. All good to go!

MinsukJi-NOAA (approved):
Looks good to me.

PR #185, opened 2020-08-12
tests/rt_cheyenne.conf: update RRTMGP regression tests

climbfuji:
## Description

This is a tiny bugfix following the RRTMGP regression test updates in the previous PR. This change affects file `tests/rt_cheyenne.conf` only and can be merged anytime, preferably as soon as possible, without having to run the tests on the tier-1 platforms.

### Issue(s) addressed

Running regression tests on cheyenne.intel using `rt_cheyenne.conf` fails because the old rrtmgp regression tests no longer exist (replaced by two new tests with different names).

## Testing

Tested on cheyenne.intel using `rt.sh -l rt_cheyenne.conf ...`

## Dependencies

n/a


climbfuji:
This has been pulled into https://github.com/ufs-community/ufs-weather-model/pull/183 and will be merged as part of it. Closing this PR.

PR #160, opened 2020-07-02
implement running a single rt test

MinsukJi-NOAA:
This PR addresses issue #159

PR #188, opened 2020-08-20
Esmf810

junwang-noaa:
## Description

-Add GFSv16 performance tuning to fv3 develop branch.
-rt_orion.conf is merged into rt.conf, and orion baseline is moved to emc.nemspara account.
The baseline results will change because the isrctermprocessing is changed from 1 to 0, this will allow a more balanced regridding between fcst grid and write grid.

### Issue(s) addressed

[ufs-community/ufs-weather-model#142](https://github.com/ufs-community/ufs-weather-model#142)

## Testing

The RT test has been run on dell, cray, orion, is now running on cray.

Intel compiler.
Are the changes covered by regression tests?
Yes.

## Dependencies

[fv3atm PR #160](https://github.com/NOAA-EMC/fv3atm/pull/160)
[nems PR #76](https://github.com/NOAA-EMC/NEMS/pull/76)

These two PRs need to be committed first.


conf/configure.fv3.wcoss_cray
```diff
@@ -48,7 +48,7 @@ CFLAGS := $(INCLUDE)
 FFLAGS := $(INCLUDE) -fno-alias -auto -safe-cray-ptr -ftz -assume byterecl -nowarn -sox -align array64byte
 
 CPPDEFS += -Duse_libMPI -Duse_netCDF -DSPMD -DUSE_LOG_DIAG_FIELD_INFO  -DUSE_GFSL63 -DGFS_PHYS -Duse_WRTCOMP
-CPPDEFS += -DNEW_TAUCTMAX -DINTERNAL_FILE_NML -DNO_PARALLEL_NETCDF
+CPPDEFS += -DNEW_TAUCTMAX -DINTERNAL_FILE_NML
```
DusanJovic-NOAA, line 51: if PARALLEL_NETCDF is now enabled on Cray, add:
`set(PARALLEL_NETCDF ON  CACHE BOOL "Enable parallel NetCDF" FORCE)`
in `cmake/configure_wcoss_cray.cmake`, similar to configure_wcoss_dell_p3.cmake, for example

climbfuji:
@junwang-noaa I am going through my 777 (!) emails while I was on leave, then look at this PR first thing.

tests/rt.sh
```diff
@@ -355,11 +355,6 @@ SINGLE_NAME=''
 
 TESTS_FILE='rt.conf'
 
-if [[ $MACHINE_ID = orion.* ]]; then
```
climbfuji, line 358: Should we do the same with Cheyenne and Stampede in the future?

climbfuji (commented):
Looks good so far, I am compiling the software stack for hera.gnu and will run the tests (including the parallel tests) afterwards.

tests/rt.sh
junwang-noaa: I think so, but I think we may need an option to define which tests we will not run on a certain platforms.

climbfuji: Yes, I remember we talked about this. Something like a `-cheyenne.intel`, i.e. the `-` negates running the test on that platform. Same for `COMPILE` and `RUN` lines. Not that for compile steps that work on all platforms, you can just use one `COMPILE` line and leave the machine column empty.

DusanJovic-NOAA (approved):

climbfuji (approved):

PR #191, opened 2020-08-27
Update modulefiles for tier-2 systems for ESMF 8.1.0 bs 21

climbfuji:
## Description

- Update modulefiles for cheyenne.{gnu,intel}, gaea.intel, jet.intel to use ESMF 8.1.0 bs21
- Changemodulefile for hera.gnu so that it matches the tier-2 modulefiles

### Issue(s) addressed

Code no longer compiles on tier-2 platforms after the ESMF 8.1.0 bs21 update was made for tier-1 platforms.

## Testing

ufs-weather-model compiles and runs on all tier-2 platforms (gaea.intel, jet.intel, cheyenne.intel, cheyenne.gnu), on endeavor.intel and also on macOS with gcc+gfortran and clang+gfortran

Regression testing:
- all tests pass on hera.gnu, hera.intel, wcoss_cray, wcoss_dell_p3, orion.intel; regression test log updated in the PR
- extra credits: all tests pass on cheyenne.gnu; regression test log updated in the PR

## Dependencies

- waiting on https://github.com/NOAA-EMC/fv3atm/pull/164


modulefiles/cheyenne.gnu/fv3
```diff
@@ -21,11 +21,13 @@ module load mpt/2.19
 module load ncarcompilers/0.5.0
 
 ##
-## use pre-compiled NetCDF, ESMF and NCEP libraries for above compiler / MPI combination
+## use pre-compiled PNG, NetCDF, EMSF and NCEP libraries for above compiler / MPI combination
```
DusanJovic-NOAA: EMSF -> ESMF

junwang-noaa:
Dom, do you have ESMF compiled in debug mode? In s2s, the debug mode test
is using the ESMF lib compiled with debug mode, we may need to update
ufs-weather debug test with that when we merging the two repoes.

climbfuji:
> Dom, do you have ESMF compiled in debug mode? In s2s, the debug mode test is using the ESMF lib compiled with debug mode, we may need to update ufs-weather debug test with that when we merging the two repoes.
> […](#)

No I haven't this wasn't necessary until now. The current NCEPLIBS-external and NCEPLIBS build do not support this functionality yet. Would we build everything (i.e. netCDF, ...) in DEBUG mode or "just" ESMF?

modulefiles/cheyenne.gnu/fv3
climbfuji: Argh. Thanks, will fix.

junwang-noaa:
That is OK, I will only add fv3 module then. So far only ESMF is built with
debug mode, it is important for s2s. I think it is good to have all the
libs built in debug mode.

climbfuji:
> That is OK, I will only add fv3 module then. So far only ESMF is built with debug mode, it is important for s2s. I think it is good to have all the libs built in debug mode.
> […](#)

If that's really needed and there is a way to switch to different "-debug" modules, then I can accommodate this for ESMF only in NCEPLIBS-external easily and provide debug modules relatively quickly. I would defer compiling the entire stack in debug mode to when we/I switch to hpc-stack. I assume hpc-stack has an option to compile everything in debug mode.

climbfuji:
@DusanJovic-NOAA @junwang-noaa this PR should be ready to test. I will run regression tests on hera.intel, hera.gnu, orion.intel.

junwang-noaa (approved):

.gitmodules
```diff
@@ -17,5 +17,4 @@
 [submodule "stochastic_physics"]
 	path = stochastic_physics
 	url = https://github.com/noaa-psd/stochastic_physics
-        branch = master
-
+	branch = master
```
climbfuji, line 20: Fixing the additional indents and newline from Clara's PR

DusanJovic-NOAA (approved):

PR #209, opened 2020-09-30
Use top-level build.sh in regression test compile scripts

DusanJovic-NOAA:
## Description

Use top-level build.sh in the regression test compile scripts.

### Issue(s) addressed

Current compile.sh script in the regression test invokes cmake/make directly. This PR changes that and runs the top-level build.sh script.

## Testing

Full regression tests will be run on all supported platform. No changes in the baseline.

## Dependencies

N/A


climbfuji (approved):
Looks good to me (I only read through the changes, didn't test them).

CMakeLists.txt
```diff
@@ -262,23 +262,23 @@ target_link_libraries(ufs-weather-model-lib PUBLIC fv3atm)
 ###############################################################################
 ### UFS executable
 ###############################################################################
-add_executable(NEMS.exe
+add_executable(ufs_model
```
junwang-noaa, line 265: Is the executable name? Can we use ufs_model.exe to avoid confusion?

DusanJovic-NOAA: In cmake context this functions declares a target, which will create an executable, and targets typically do not have extension. Another reason is that I think we are moving away from using the .exe suffix (f.e. ufs_utils executables)

DusanJovic-NOAA: @aerorahul Do you have an opinion or suggestion about this, since I know you will be updating all install targets in s2s version of this file.

junwang-noaa: We may need more discussion, let's take it as is for now.

junwang-noaa (approved):

CMakeLists.txt
aerorahul: executables don't have an extension. 
The executable name should reflect the project name.
ufs_model is fine for now for the executable.

build.sh
```diff
@@ -14,14 +14,12 @@ export CMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER:-mpif90}
...
 
-CCPP_SUITES="${CCPP_SUITES:-FV3_GFS_v15p2}"
-CMAKE_FLAGS+=" -DCCPP_SUITES=${CCPP_SUITES} -DNETCDF_DIR=${NETCDF}"
+[[ -n "${CCPP_SUITES:-""}" ]] && CMAKE_FLAGS+=" -DCCPP_SUITES=${CCPP_SUITES}"
+CMAKE_FLAGS+=" -DNETCDF_DIR=${NETCDF}"
 
 cd ${BUILD_DIR}
-cmake .. ${CMAKE_FLAGS}
+cmake ${UFS_MODEL_DIR} ${CMAKE_FLAGS}
 make -j ${BUILD_JOBS:-4}
```
aerorahul, line 25: can we add verbose option, default off?
e.g.
`make -j ${BUILD_JOBS:-4} VERBOSE=${BUILD_VERBOSE:-0}`

aerorahul (approved):
I just have one comment, but other than that, everything looks good to me.

CMakeLists.txt
junwang-noaa: @aerorahul, "executables don't have an extension." seems just cmake convention, if we put it in a context of test system, giving project name as executable name is confusing as it's hard to know what the file is, real project name or an executable. But this is just my opinion. I see the executable is copied to ${BUILD_NAME}.exe, so I am find  for it in rt.sh

build.sh
DusanJovic-NOAA: Good idea. Let's add this next time, to avoid rerunning all the tests. And BUILD_VERBOSE should be set/exported in compile.sh.

climbfuji: > Good idea. Let's add this next time, to avoid rerunning all the tests. And BUILD_VERBOSE should be set/exported in compile.sh.

This is a great idea. - next commit!

CMakeLists.txt
aerorahul: what do you mean cmake convention? CMake is not enforcing anything here on the name.  You could name it whatever.
Do you see a `ls.exe` or `grep.exe` or `nc-config.x` or whatever you see in `/usr/bin` or any of the compiled executables such as `ncdump`, `h5dump` or `icc.exe`?

The choice of `.exe` comes from Windows.  That is a convention.
The convention on Linux and Unix is, executables do not have an extension.  They are determined by the permissions on the file.

junwang-noaa:
Rahul, The question is if I see " nc-config", I don't know if it is a
configuration file or a utility executable, but if I see nc-config.x, I
know it's an executable.  I don't mind not having the convention of .exe,
but in general we'd like to avoid confusion so that it's better that the
name itself tells about how it will be used. "ufs-model" can be used in
cmake context, out of that context, I don't know what the name ufs-model
tells developers, a project, a model repo, a documentation, a configuration
file, etc.

aerorahul:
@junwang-noaa 
When the package is installed (via `make install`), all executables are located in the package's `bin/` directory. (i.e. if the build system is written properly)
It should not manually moved or copied to a "convenient" location.
That is the source of confusion.  We move compiled executables around, and to distinguish that, we add our conceptions of how to eliminate confusion.
If the executable is built and installed (`make install`) (in `bin/`) there is no doubt that the file is an executable.
The build script is missing this last step.

junwang-noaa:
Rahul, I'd think this is a readability issue. Developers should be free to
copy executable from where it is built, say another machine (e.g., run hera
model executable on orion). So someone copied the ufs-model to another
machine (in a canned case, or a directory with many other files), people
may confuse what it is.

aerorahul:
No one should be copying executables from one machine to another. Period.
There is no good reason for doing that.

Think about what we are enabling by doing that?

And even if someone did want to do that,  why can't they copy from `bin/` and place it wherever they want and name it whatever they want?

It is not a readability issue.  It is about doing things properly. 
What is confusing about this:
```
ufs_install_prefix/
├── bin
│   └── ufs_model
├── include/
 |    ├──FV3/
 |    ├── ...
├── lib
     ├── libfv3atm.a
     ├── ...
     ├── libccpp.a
     └── cmake/
```

junwang-noaa:
I see your point. But we are talking about two different things.  you are
saying about build system (no confusion I agree), but I am talking about
applications.

aerorahul:
Yes, an application e.g. `rt.sh` can do whatever it wants with the executable.  I can copy it (not move), rename it to `fv3_1.exe` and so on.  The original executable should not leave the `bin/` directory after install (except `make clean`).
Application in ops can call it `gfs.x` or whatever.
Although, renaming it to something other than what was built is a source of confusion too.
What an application does is different than the outcome of the build system of the package.

climbfuji:
The only issue I would like to raise is that we are making a lot of changes in a very short amount of time.

Developers using rt.sh are used to finding NEMS.x, which is then moved to tests/fv3.exe and then copied or linked to the run directory. Developers using the global workflow are used to something else.

Whatever we change needs to be communicated very clearly, and maybe it is a good idea to combine certain changes or slow down certain developments to give people a chance to adapt.

It's not a question of what is right and what is wrong, it's about gettting to where we want to get without frustrating the users and developers.

PR #212, opened 2020-10-03
GFS.v16: using NCO installed upp lib on wcoss

junwang-noaa:
## Description

1. NCO changed post lib name to upp lib on wcoss for GFS.v16 implementation. Model needs to use UPP related lib name and module include path.
2. Post control files have been updated with decimal scale change, the post control files need to be update to be consistent with the ones in POST repo.

## Testing

Code has been tested with fv3 parallel test case.

## Dependencies

[fv3atm PR#182](https://github.com/NOAA-EMC/fv3atm/pull/182)


WenMeng-NOAA (approved):
looks good to me.

PR #213, opened 2020-10-05
Preparation for S2S cmake

aerorahul:
## Description

In this PR:

- All `#ifdef OPENMP` are replaced with `#ifdef _OPENMP` to follow standard compiler macro for OpenMP.
- CCPP code generation is moved inside of `FV3/`.  This will be redundant when the Data Atmosphere component comes in. Updates to the python script was performed to remove the hard-wired `FV3/` path name in the file.
- `cmake/GNU.cmake` and `cmake/Intel.cmake` hard-wire global compiler definitions with `add_definitions`.  These are now applied with `target_compile_definitions`. 
- CMakeLists.txt is passed through cmake-norm checker.
- `INSTALL_INTERFACE` is added, so that this component can be installed and linked in downstream applications e.g. JEDI.
This may be incomplete, so will need more work.  
 
What bug does it fix, or what feature does it add?  
Significantly cleans up the CMake build and prepares for S2S and Data Atmosphere integration.

Is a change of answers expected from this PR?  
No.

## Testing

Regression tests.
All standard suite of compilers and HPC's on NOAA Tier-1 Platforms
Are the changes covered by regression tests? 
Yes.
Have regression tests and unit tests (utests) been run? On which platforms and with which compilers?
Yes. On NOAA Tier-1 platforms.

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
Those branches should have matching names (ideally)
fv3atm:: [feature/overload](https://github.com/aerorahul/fv3atm/tree/feature/overload)
stochastic_physics:: [feature/overload](https://github.com/aerorahul/stochastic_physics/tree/feature/overload)

Do PRs in upstream repositories need to be merged first? Yes
fv3atm: https://github.com/NOAA-EMC/fv3atm/pull/183
stochastic physics: https://github.com/noaa-psd/stochastic_physics/pull/30


aerorahul:
All NOAA Tier-1 Regression Tests have been run and logs uploaded.

climbfuji (approved):
I compiled the code on my macOS laptop with clang+gfortran in DEBUG mode, and with gcc+gfortran in PROD mode. The executables were created as expected. And it's still called `fv3.exe` as one would expect it, yeah :-)

junwang-noaa (approved):

DusanJovic-NOAA (approved):

PR #220, opened 2020-10-15
GFSv16: update GFS.v16 library version on orion and hera 

junwang-noaa:
## Description

This PR will update the hera/orion module files to use the GFS.v16 prod libraries:
g2/3.2.0
crtm/2.3.0
w3emc/2.4.0
w3nco/2.2.0
upp/8.0.11

No results will change from this PR. GFS.v16.0.13 only has prod lib change on wcoss( these prod lib were not available on hera/orion at that time). After this PR, we will be able to run the final gfs.v16 implementation package on wcoss, hera and orion.

## Testing

The code was tested with a RT test case and a fv3 parallel test case on hera/orion/wcoss.

## Dependencies

No.

### DIFF
--- a/conf/configure.fv3.hera.intel
+++ b/conf/configure.fv3.hera.intel
@@ -37,7 +37,7 @@ include       $(ESMFMKFILE)
 ESMF_INC    = $(ESMF_F90COMPILEPATHS)
 
 NEMSIOINC = -I$(NEMSIO_INC)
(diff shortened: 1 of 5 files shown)


WenMeng-NOAA (approved):
It looks good to me.

yangfanglin (approved):
Looks good, although I did not check every line of changes

Issue #242, opened 2020-10-25
alternate comparison method needed for mediator restart files


DeniseWorthen:
## Description
In the current RT, the comparison of the mediator restart file will sometimes fail, even though all other files pass comparison. In these cases, manual comparison of the mediatory history files will show that the actual data in the file is identical. 

## Solution
When the mediator history file is the only file to fail comparison, we need a second comparison method for the mediator restart file in the RT system.

MinsukJi-NOAA:
#261 addressing this issue has been merged into develop

PR #261, opened 2020-10-31
rt update

MinsukJi-NOAA:
## Description

Update rt test to:
- use compare_ncfile.py if cmp fails for mediator restart files (#242) or netcdf_parallel files
- include cpld 35d bmark tests
- combine with PR #273 

## Testing

RT will be performed on tier 1 platforms.

## Dependencies

N/A


DeniseWorthen:
This looks like it should work for DATM also, as long as there was an rt.conf file for that use case, correct?

MinsukJi-NOAA:
> This looks like it should work for DATM also, as long as there was an rt.conf file for that use case, correct?

Yes, you are right.

DeniseWorthen:
I got an error when trying the 35d no wave test using ecflow using ``./rt.sh -fek -l rt_35d.conf >output 2>&1 &``

My test build directory is here: /scratch2/NCEPDEV/climate/Denise.Worthen/WORK/ufs_rt/tests

The only change I made was to edit the test to run only 2 days instead of 35d.

MinsukJi-NOAA:
@DeniseWorthen, are you using ecFlow and using your own executable? Not sure if that works...

DeniseWorthen:
I did not try to build the executable first. This was just checking out your branch, editing the tests/cpld_bmark_35d for 2d instead of 35d and then launching the rt_35d.conf, including the build. 

I always ran with ecflow when doing the 35d tests in ufs-s2s-model.

MinsukJi-NOAA:
> I did not try to build the executable first. This was just checking out your branch, editing the tests/cpld_bmark_35d for 2d instead of 35d and then launching the rt_35d.conf, including the build.
> 
> I always ran with ecflow when doing the 35d tests in ufs-s2s-model.

Just realized that the 35d bmark tests are run on Orion only.  35 days with and without wave were tested on Orion for the full 35d period. Your error message on Hera seems to be related to the ecflow though.

DeniseWorthen:
The 35d tests were restricted to orion when we had lost access to hera for the prototype runs. But that restriction is now lifted so I think we should be able to run them on either hera or orion. 

But you are right, the current rt_35d.conf is restricted to compiling on orion. I added hera.intel and it appears to have resolved the first failure. I don't think the baselines have all the inputs they need on hera though.

DeniseWorthen:
Actually the BM_IC does exist on hera. I've verified that the 35d test (edited for 2d) run there as long as hera.intel is added to the compile line.

tests/rt_35d.conf
```diff
@@ -0,0 +1,9 @@
+COMPILE | CCPP=Y SUITES=FV3_GFS_2017_coupled,FV3_GFS_2017_satmedmf_coupled,FV3_GFS_v15p2_coupled S2S=Y  |   | orion.intel | fv3 | |
```
DeniseWorthen: Can we add hera.intel here and in the rt_wave_35d.conf?

MinsukJi-NOAA: Will do. How about without wave?

DeniseWorthen: The wave input is also present in the BM_IC on hera. These are just links to the actual BM input except for the 20130401 date.

DeniseWorthen (approved):

tests/rt.sh
```diff
@@ -207,6 +223,7 @@ elif [[ $MACHINE_ID = hera.* ]]; then
   export PYTHONPATH=/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/ecflow/lib/python2.7/site-packages
   ECFLOW_START=/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/ecflow/bin/ecflow_start.sh
   ECF_PORT=$(( $(id -u) + 1500 ))
+  NCCMP='/scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/v1.0.0-beta1/intel-18.0.5.274/impi-2018.0.4/nccmp/1.8.7.0/bin/nccmp'
```
aerorahul: please don't hardwire paths.  replace with:`NCCMP=$(which nccmp)`
The modulefile for nccmp puts the executable in PATH.

tests/rt.sh
```diff
@@ -238,6 +255,8 @@ elif [[ $MACHINE_ID = orion.* ]]; then
   export PYTHONPATH=/work/noaa/fv3-cam/djovic/ecflow/lib/python2.7/site-packages
   ECFLOW_START=/work/noaa/fv3-cam/djovic/ecflow/bin/ecflow_start.sh
   ECF_PORT=$(( $(id -u) + 1500 ))
+  NCCMP='/apps/contrib/NCEP/libs/hpc-stack/v1.0.0-beta1/intel-2018.4/impi-2018.4/nccmp/1.8.7.0/bin/nccmp'
```
aerorahul: Same comment as above.

tests/utest
```diff
@@ -475,6 +477,7 @@ elif [[ $MACHINE_ID = wcoss_dell_p3 ]]; then
   STMP=/gpfs/dell2/stmp
   PTMP=/gpfs/dell2/ptmp
   SCHEDULER=lsf
+  NCCMP=''
```
aerorahul: There is a `nccmp` module on wcoss_dell_p3.

tests/utest
```diff
@@ -447,8 +449,8 @@ elif [[ $MACHINE_ID = orion.* ]]; then
   DISKNM=/work/noaa/nems/emc.nemspara/RT
   STMP=$dprefix/stmp
   PTMP=$dprefix/stmp
-
   SCHEDULER=slurm
+  NCCMP='/apps/contrib/NCEP/libs/hpc-stack/v1.0.0-beta1/intel-2018.4/impi-2018.4/nccmp/1.8.7.0/bin/nccmp'
```
aerorahul: please don't hardwire paths.  replace with:`NCCMP=$(which nccmp)`
The modulefile for nccmp puts the executable in PATH.

tests/utest
```diff
@@ -424,6 +425,7 @@ if [[ $MACHINE_ID = hera.* ]]; then
   STMP=${dprefix}/stmp4
   PTMP=${dprefix}/stmp2
   SCHEDULER=slurm
+  NCCMP='/scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/v1.0.0-beta1/intel-18.0.5.274/impi-2018.0.4/nccmp/1.8.7.0/bin/nccmp'
```
aerorahul: Same comment as below.

tests/rt.sh
```diff
@@ -56,6 +56,19 @@ rt_single() {
...
 
+rt_35d() {
+  local sy=$(echo ${DATE_35D} | cut -c 1-4)
+  local sm=$(echo ${DATE_35D} | cut -c 5-6)
+  local new_test_name="tests/${TEST_NAME}_${DATE_35D}"
+  rm -f $new_test_name
+  cp tests/$TEST_NAME $new_test_name
+
+    sed -i -e "s/\(export SYEAR\)/\1=\"$sy\"/" $new_test_name
+    sed -i -e "s/\(export SMONTH\)/\1=\"$sm\"/" $new_test_name
```
aerorahul: unintended indents.

aerorahul (changes requested):
see comments about hard-wired paths

junwang-noaa:
Minsuk, I see the nccmp is only applied to the coupled restart file, is it possible that we can apply it to all the nc files.  Currently in the RT, I commented out the two parallel netcdf tests (regional one is in my PR#206)  whose output files need to be compared with nccmp, and cmp comparison will show a failure. So to test then, please uncomment the two tests in rt.conf:

#RUN     | fv3_ccpp_regional_quilt_netcdf_parallel                                                                                        | standard    |                | fv3         |
and
#RUN     | fv3_ccpp_wrtGauss_netcdf_parallel                                                                                              | standard    |                | fv3         |

MinsukJi-NOAA:
> Minsuk, I see the nccmp is only applied to the coupled restart file, is it possible that we can apply it to all the nc files. Currently in the RT, I commented out the two parallel netcdf tests (regional one is in my PR#206) whose output files need to be compared with nccmp, and cmp comparison will show a failure. So to test then, please uncomment the two tests in rt.conf:
> 
> #RUN | fv3_ccpp_regional_quilt_netcdf_parallel | standard | | fv3 |
> and
> #RUN | fv3_ccpp_wrtGauss_netcdf_parallel | standard | | fv3 |

fv3_ccpp_wrtGauss_netcdf_parallel_prod is still not able to pass.

```
baseline` dir = /work/noaa/nems/emc.nemspara/RT/NEMSfv3gfs/develop-20201022/INTEL/fv3_wrtGauss_netcdf_ccpp
working dir  = /work/noaa/stmp/jminsuk/stmp/jminsuk/FV3_RT/rt_1077/fv3_ccpp_wrtGauss_netcdf_parallel_prod
Checking test 002 fv3_ccpp_wrtGauss_netcdf_parallel results ....
 Comparing atmos_4xdaily.tile1.nc .........OK
 Comparing atmos_4xdaily.tile2.nc .........OK
 Comparing atmos_4xdaily.tile3.nc .........OK
 Comparing atmos_4xdaily.tile4.nc .........OK
 Comparing atmos_4xdaily.tile5.nc .........OK
 Comparing atmos_4xdaily.tile6.nc .........OK
 Comparing phyf000.nc ............NOT OK...TRY NCCMP.......NOT OK
 Comparing phyf024.nc ............NOT OK...TRY NCCMP.......NOT OK
 Comparing dynf000.nc ............NOT OK...TRY NCCMP.......NOT OK
 Comparing dynf024.nc ............NOT OK...TRY NCCMP.......NOT OK
```
```
baseline dir = /work/noaa/nems/emc.nemspara/RT/NEMSfv3gfs/develop-20201022/INTEL/fv3_regional_quilt_netcdf_parallel_ccpp
working dir  = /work/noaa/stmp/jminsuk/stmp/jminsuk/FV3_RT/rt_1077/fv3_ccpp_regional_quilt_netcdf_parallel_prod
Checking test 007 fv3_ccpp_regional_quilt_netcdf_parallel results ....
 Comparing atmos_4xdaily.nc .........OK
 Comparing dynf000.nc .........OK
 Comparing dynf024.nc ............NOT OK...TRY NCCMP......OK
 Comparing phyf000.nc .........OK
 Comparing phyf024.nc ............NOT OK...TRY NCCMP......OK
Test 007 fv3_ccpp_regional_quilt_netcdf_parallel PASS
```

junwang-noaa:
Minsuk, Thanks for testing it. I realized one bug in
fv3_ccpp_wrtGauss_netcdf_parallel_prod. Can you change the following line
in tests/tests/fv3_ccpp_wrtGauss_netcdf_parallel
<https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/tests/fv3_ccpp_wrtGauss_netcdf_parallel>

export CNTL_DIR=fv3_wrtGauss_netcdf
to
export CNTL_DIR=fv3_wrtGauss_netcdf_parallel

MinsukJi-NOAA:
> Minsuk, Thanks for testing it. I realized one bug in fv3_ccpp_wrtGauss_netcdf_parallel_prod. Can you change the following line in tests/tests/fv3_ccpp_wrtGauss_netcdf_parallel <https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/tests/fv3_ccpp_wrtGauss_netcdf_parallel> export CNTL_DIR=fv3_wrtGauss_netcdf to export CNTL_DIR=fv3_wrtGauss_netcdf_parallel
> […](#)

Jun, that change made the test pass.  I included the change in my latest commit.

junwang-noaa:
Thanks, please turn on the two tests in your commit. -Jun

aerorahul (approved):
Thank you for addressing comments.

tests/rt_35d.conf
```diff
@@ -0,0 +1,9 @@
+COMPILE | CCPP=Y SUITES=FV3_GFS_2017_coupled,FV3_GFS_2017_satmedmf_coupled,FV3_GFS_v15p2_coupled S2S=Y  |   | hera.intel orion.intel | fv3 | |
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2012010100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2012040100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2012070100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2012100100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2013010100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2013040100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2013070100
+RUN     | cpld_bmark_35d |   |             | fv3 |  | 2013100100
```
DusanJovic-NOAA, line 9: Why are there 8 35d tests? Is one 35d test run not sufficient?

tests/rt_wave_35d.conf
```diff
@@ -0,0 +1,9 @@
+COMPILE | CCPP=Y SUITES=FV3_GFS_2017_coupled,FV3_GFS_2017_satmedmf_coupled,FV3_GFS_v15p2_coupled S2S=Y WW3=Y  |   | hera.intel orion.intel | fv3 | |
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2012010100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2012040100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2012070100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2012100100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2013010100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2013040100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2013070100
+RUN     | cpld_bmark_wave_35d |   |             | fv3 |  | 2013100100
```
DusanJovic-NOAA: Can this be combined with rt_35d.conf? Why is this a separate .conf file?

MinsukJi-NOAA:
I did some comparison between nccmp and compare_ncfile.py. nccmp is okay with mediator restart files, but takes too long for netcdf_parallel files.
netcdf

aerorahul:
I think there is something wrong with `nccmp` or its use, or the netCDF file.  Compression perhaps?
Otherwise why is a `97M` file `dynf024.nc` taking 89m while a `2G` file `ufs.cpld.cpl` file taking 2s.

MinsukJi-NOAA:
> I think there is something wrong with `nccmp` or its use, or the netCDF file. Compression perhaps?
> Otherwise why is a `97M` file `dynf024.nc` taking 89m while a `2G` file `ufs.cpld.cpl` file taking 2s.

I agree.  Can't really make sense of the results. Maybe I don't have sufficient knowledge of the netcdf.

junwang-noaa:
Yes, the parallel netcdf test cases output compressed netcdf data. Maybe we
can use compare_ncfile.py since it only takes seconds for the comparison.

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
...
+
+nc1 = Dataset(sys.argv[1])
+nc2 = Dataset(sys.argv[2])
+for varname in nc1.variables.keys():
+    data1 = nc1[varname][:]
+    data2 = nc2[varname][:]
+    diff = data2-data1
+    if (np.abs(diff)).max() != 0:
+      print("The two nc files are different.")
+      break
```
aerorahul: This should exit with a return code of non-zero.
replace with `sys.exit(1)`

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+import os
+import sys 
+import re
+import numpy as np
+##from __future__ import print_function
+from netCDF4 import Dataset
+
+nc1 = Dataset(sys.argv[1])
+nc2 = Dataset(sys.argv[2])
```
aerorahul: You open, but you don't close.

Please replace with:
```
with Dataset(sys.argv[1]) as nc1, Dataset(sys.argv[2]) as nc2:
```

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+import os
```
aerorahul: not used.

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
...
+import sys 
+import re
```
aerorahul: not used.

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
...
+from netCDF4 import Dataset
+
+nc1 = Dataset(sys.argv[1])
+nc2 = Dataset(sys.argv[2])
+for varname in nc1.variables.keys():
+    data1 = nc1[varname][:]
+    data2 = nc2[varname][:]
+    diff = data2-data1
+    if (np.abs(diff)).max() != 0:
+      print("The two nc files are different.")
```
aerorahul: why not print which variable has non-zero differences.
It would be a bit more meaningful.

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+import os
+import sys 
+import re
+import numpy as np
+##from __future__ import print_function
```
aerorahul: You are not requiring python3 in the interpreter.  This line is necessary if `/usr/bin/env python` returns python version 2.

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
...
+from netCDF4 import Dataset
+
+nc1 = Dataset(sys.argv[1])
+nc2 = Dataset(sys.argv[2])
+for varname in nc1.variables.keys():
```
aerorahul: Shouldn't we first check if the variables in the files are the same?  What if variables in either file are missing or added?

tests/compare_ncfile.py
```diff
@@ -0,0 +1,17 @@
...
+    data1 = nc1[varname][:]
```
aerorahul: what is the timing on these?  I think you can speed it up further if you don't copy into local variables.  Especially for large arrays.

aerorahul (changes requested):
Some comments for the new addition of the python script.

modulefiles/hera.gnu/fv3
```diff
@@ -40,3 +40,5 @@ module load nemsio/2.5.2
 module load sp/2.3.3
 module load w3emc/2.7.3
 module load w3nco/2.4.1
+
```
DusanJovic-NOAA: If nccmp is not used anymore, remove this module.

tests/rt.sh
```diff
@@ -117,6 +130,7 @@ if [[ $MACHINE_ID = wcoss_cray ]]; then
   module load ecflow/intel/4.7.1
   ECFLOW_START=${ECF_ROOT}/intel/bin/ecflow_start.sh
   ECF_PORT=$(grep $USER /usrx/local/sys/ecflow/assigned_ports.txt | awk '{print $2}')
+  NCCMP=''
```
DusanJovic-NOAA: If nccmp is not used anymore, remove this variable.

tests/rt_utils.sh
```diff
@@ -270,9 +270,25 @@ check_results() {
         d=$( cmp ${RTPWD}/${CNTL_DIR}/$i ${RUNDIR}/$i | wc -l )
 
         if [[ $d -ne 0 ]] ; then
-          echo ".......NOT OK" >> ${REGRESSIONTEST_LOG}
-          echo ".......NOT OK"
-          test_status='FAIL'
+
+          if [[ ${MACHINE_ID} =~ orion || ${MACHINE_ID} =~ hera ]]; then
+            printf ".......NOT OK...TRY ALT CHECK.." >> ${REGRESSIONTEST_LOG}
+            printf ".......NOT OK...TRY ALT CHECK"
```
aerorahul: We should not have a "NOT OK" and then and "OK" in the same log file.
There will be a "NOT OK" or an "OK"

tests/rt_utils.sh
```diff
@@ -270,9 +270,25 @@ check_results() {
...
+            else
+              echo "....OK" >> ${REGRESSIONTEST_LOG}
+              echo "....OK"
+            fi
+          else
+            echo ".......NOT OK" >> ${REGRESSIONTEST_LOG}
+            echo ".......NOT OK"
+            test_status='FAIL'
+          fi
+
```
aerorahul: There is too much logic of OK/NOT OK going on here.
It can be cleaned up to a single OK/NOT OK message.

junwang-noaa:
Can we just have "ALT CHECK OK" if compare_ncfile.py is used?

tests/rt_utils.sh
aerorahul: This whole section from Lines 272 to  295 can be boiled down to this:
```
 if [[ $d -ne 0 ]] ; then
  if [[ ${MACHINE_ID} =~ orion || ${MACHINE_ID} =~ hera ]]; then
    d=$( ${PATHRT}/compare_ncfile.py ${RTPWD}/${CNTL_DIR}/$i ${RUNDIR}/$i 2>/dev/null | wc -l )
  fi
fi

if [[ $d -ne 0 ]]; then
  echo ".....NOT OK" >> ${REGRESSIONTEST_LOG}
  echo ".....NOT OK"
  test_status='FAIL'
else
  echo "....OK" >> ${REGRESSIONTEST_LOG}
  echo "....OK"
fi
```

junwang-noaa:
Does compare_ncfile.py compare the meta data and global attribute too
besides data fields?

tests/rt_utils.sh
```diff
@@ -270,9 +270,25 @@ check_results() {
...
 
         if [[ $d -ne 0 ]] ; then
-          echo ".......NOT OK" >> ${REGRESSIONTEST_LOG}
-          echo ".......NOT OK"
-          test_status='FAIL'
+
+          if [[ ${MACHINE_ID} =~ orion || ${MACHINE_ID} =~ hera ]]; then
+            printf ".......NOT OK...TRY ALT CHECK.." >> ${REGRESSIONTEST_LOG}
+            printf ".......NOT OK...TRY ALT CHECK"
+            d=$( ${PATHRT}/compare_ncfile.py ${RTPWD}/${CNTL_DIR}/$i ${RUNDIR}/$i 2>/dev/null | wc -l )
```
aerorahul, line 276: Just noticed this.
Why is a `wc -l` being done here.  One can operate on the return code of this script.
This can be replaced by:
```
python ${PATHRT}/compare_ncfile.py ${RTPWD}/${CNTL_DIR}/$i ${RUNDIR}/$i 
d=$?
```

MinsukJi-NOAA: That won't work. Because of set -e, the script will exit. Can try piping the two lines.

MinsukJi-NOAA:
> Can we just have "ALT CHECK OK" if compare_ncfile.py is used?

Done

MinsukJi-NOAA:
> Does compare_ncfile.py compare the meta data and global attribute too besides data fields?

It does not compare the metadata, and global attributes. It only compares the data fields for all variables (not attributes here, either).

tests/rt_wave_35d.conf
MinsukJi-NOAA: @DeniseWorthen, Is combining rt_35d.conf with rt_wave_35d.conf something that we would like to pursue?

tests/rt_35d.conf
DeniseWorthen: We've found that multiple ICs are required to give a good sense of whether the coupled model is stable.

tests/rt_wave_35d.conf
DeniseWorthen: I have no real objection to combining these. We generally have not been testing 16 cases (8 w/ wave and 8 w/o), but a person could easily comment out the tests they wished to skip.

tests/rt_35d.conf
aerorahul: So long as this does not become a requirement to run with every PR, add 800 or 8000.
What is the guarantee that the model won't fail on Day 36 or on Day 1 of any other initial condition date.

These are ad hoc tests, and adding ad hoc tests complicates `rt.sh`

MinsukJi-NOAA:
@DusanJovic-NOAA, I am getting an error when creating a baseline for fv3_ccpp_regional_quilt_netcdf_parallel:
`++ rsync -arv /work/noaa/nems/emc.nemspara/RT/NEMSfv3gfs/develop-20201106/INTEL/fv3_regional_quilt_netcdf_parallel/model_configure .
sending incremental file list
rsync: change_dir "/work/noaa/nems/emc.nemspara/RT/NEMSfv3gfs/develop-20201106/INTEL/fv3_regional_quilt_netcdf_parallel" failed: No such file or directory (2)`

Requiring that an old baseline exists when generating a new baseline does not seem to be a good idea.

junwang-noaa:
Minsuk, maybe you can first copy the three directories
from develop-20201022 baseline to baseline develop-20201106..

DusanJovic-NOAA:
I hope we can cleanup this kind of dependencies during big tests cleanup, after IPD is removed and after all regional tests are updated.

tests/rt_35d.conf
junwang-noaa: These are not ad hoc tests, they are minimal stability tests to support benchmark runs. Current benchmark tests run 35 days, these tests give us basic idea if the model can be used for the benchmark. We don't want to start the full benchmark tests but find out that model can't run up to 35 days, that would be a huge waste of resources.

DeniseWorthen:
Are we ready to commit this? Shan is asking about running the 35d tests for frac/non-frac over the weekend and she would need this commit to be able to do that.

tests/rt.sh
```diff
@@ -203,9 +217,9 @@ elif [[ $MACHINE_ID = hera.* ]]; then
   ROCOTOCOMPLETE=$(which rocotocomplete)
   ROCOTO_SCHEDULER=slurm
 
-  export PATH=/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/ecflow/bin:$PATH
-  export PYTHONPATH=/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/ecflow/lib/python2.7/site-packages
-  ECFLOW_START=/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/ecflow/bin/ecflow_start.sh
+  export PATH=/scratch2/NCEPDEV/stmp3/Minsuk.Ji/opt/miniconda3/bin:$PATH
```
DusanJovic-NOAA: Is this going to be 'permanent' location of miniconda installation? In stmp3?

MinsukJi-NOAA: Yes, that is the intention. Do you foresee any issues?

DusanJovic-NOAA: Maybe, if they start scrubbing stmp on daily/weekly basis. I do not think software should be installed in /scratch2/NCEPDEV/stmp*

MinsukJi-NOAA: I only have access to stmp on both Hera and Orion. Can you cp -r miniconda3 to fv3-cam/Dusan.Jovic on both Hera and Orion?

DusanJovic-NOAA: Ok. I see. I can make a copy but I do not think my personal directory is good place for this. Let's keep it as it is until we find more 'permanent' place, f.e. /scratch1/NCEPDEV/nems/emc.nemspara/soft

climbfuji: Yes please, let's move away from any code, executables etc in user directories (e.g. Dusan's ecflow installations). I installed ecflow and all other software in shared project space so that it doesn't depend on me or my account.

Have you tried using the GSL miniconda3 module in contrib? It's all there and ready to use, no need to repeat.

MinsukJi-NOAA: I am not able to find `ecflow.so` and `ecflow_start.sh` in subdirectories of `/contrib/miniconda3/4.5.12`.

climbfuji: Got it - thanks for the info.

aerorahul (approved):

tests/rt.sh
junwang-noaa: Dusan/Minsuk, please install the ecflow related code in under emc.nemspara/soft on all the platforms. We can point to them when they are ready (maybe not in this commit if it takes time)

MinsukJi-NOAA: Done

MinsukJi-NOAA:
This PR is good to go.  CI issue is not addressed in this PR because of the [hpc-stack issue](https://github.com/NOAA-EMC/hpc-stack/issues/71)

climbfuji (approved):

junwang-noaa (approved):

DusanJovic-NOAA (approved):

Issue #235, opened 2020-10-23
Issue running both Intel and GNU regression tests using ecflow on the same machine (node)


climbfuji:
## Description

Under special circumstances, the regression tests fail for the GNU or Intel compiler in the compilation.

### To Reproduce:

- regression testing using ecflow for compiler A is kicked off on a node and then terminated using ctrl+c
- this allows the ecflow server survive, and it remembers the modules for compiler A
- on the same node (different shell or changing `RT_COMPILER` or logging out and back in), kick off the regression tests using ecflow for compiler B
- the compile jobs will fail with module load conflicts

## Output

Example:
```
...
+ [[ hera.gnu == cheyenne.* ]]
+ [[ hera.gnu == wcoss_dell_p3 ]]
+ BUILD_JOBS=8
+ hostname
+ set +x
Lmod has detected the following error: Cannot load module "hpc-gnu/9.2.0"
because these module(s) are loaded:
   hpc-intel

While processing the following module(s):
    Module fullname  Module Filename
    ---------------  ---------------
    hpc-gnu/9.2.0    /scratch2/NCEPDEV/nwprod/hpc-stack/libs/hpc-stack/v1.0.0-beta1/modulefiles/core/hpc-gnu/9.2.0.lua
    fv3              /scratch1/BMC/gmtb/Dom.Heinzeller/ufs-weather-model/ufs-weather-model-hafs-feature-hwrf-physics-20200922-update-from-authoritative-repositories/gnu/modulefiles/hera.gnu/fv3
```

climbfuji:
@DusanJovic-NOAA can you look at this when you get a chance, please? Since you implemented the ecflow-based regression testing and the reuse of an existing ecflow server, you know this system best.

junwang-noaa:
Dusan is on leave, so let me assign the issue to him so that he can work on
it when he is back.

Jun

DusanJovic-NOAA:
@climbfuji I can not reproduce the bug. I ran `rt.sh -e` on Hera and after few minutes killed it using ctrl-c. It stopped ecflow server:

```
+ echo 'ecflow tasks remaining: 93'
ecflow tasks remaining: 93
+ /scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/ufs/develop/ufs-weather-model/tests/abort_dep_tasks.py
+ [[ 93 -ne 0 ]]
+ wait 15570
+ sleep 10
^C++ echo 'rt.sh interrupted'
[...]

Issue #23, opened 2019-12-30
Regression tests that are not looking for specific output files can fail silently


climbfuji:
Some of the DEBUG regression tests do not check for the presence of output files generated by the model, the idea being that the code should just run and complete without an error. However, even if the code crashes rt.sh still thinks that the model ran ok (see issue https://github.com/ufs-community/ufs-weather-model/issues/24 for an example).

climbfuji:
This issue is avoided for the ufs_public_release branches only in https://github.com/ufs-community/ufs-weather-model/pull/25 by explicitly checking for output files from the DEBUG tests.

junwang-noaa:
Dom,

Do you still want to work on this issue, or if we can close it?

climbfuji:
I think this needs to be fixed for the develop branch - all debug tests should look for specific output files.

junwang-noaa:
I see, will add output files for debug tests in  the next develop branch
commit.

climbfuji:
@junwang-noaa do you know if this has been fixed, i.e. are all DEBUG tests looking for output files nowadays? I know that all the DEBUG tests I put in do.

junwang-noaa:
Sorry, not yet. We still need to add output files
for fv3_ccpp_control_debug and fv3_ccpp_stretched_nest_debug.

climbfuji:
Closed via https://github.com/ufs-community/ufs-weather-model/commit/248b609b7a3ba8e0ef05085218a0e3c0cd2b739c

PR #273, opened 2020-11-12
Fix running both Intel and GNU regression tests using ecflow on the same machine (node)

DusanJovic-NOAA:
## Description

Provide list of files to be compared in two debug regression test (#23).
Do not load fv3 modules in the rt.sh. The fv3 module will be loaded in the compile and run scripts (#235).
Rocoto must pass PATHTR environment variable via workflow xml file.
When rocoto is used increase build time on wcoss_dell to 60min (one of the coupled compilation times out)

### Issue(s) addressed

Issue #23 
Issue #235 

## Testing

New baselines required for fv3_ccpp_control_debug and fv3_ccpp_stretched_nest_debug.

### DIFF
--- a/tests/rt.sh
+++ b/tests/rt.sh
@@ -99,12 +99,7 @@ source rt_utils.sh
 
 if [[ $MACHINE_ID = wcoss_cray ]]; then
 
-  source $PATHTR/NEMS/src/conf/module-setup.sh.inc
   module load xt-lsfhpc
-
-  module use $PATHTR/modulefiles/${MACHINE_ID}
-  module load fv3
-
   module load python/2.7.14
 
   module use /usrx/local/emc_rocoto/modulefiles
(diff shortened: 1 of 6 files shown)


climbfuji:
Thanks for fixing those debug tests! Just curious, does this still compile without the `-e` or `-c` flags?

DusanJovic-NOAA:
> Thanks for fixing those debug tests! Just curious, does this still compile without the `-e` or `-c` flags?

Yes, it does.

tests/rt.sh
```diff
@@ -99,12 +99,7 @@ source rt_utils.sh
 
 if [[ $MACHINE_ID = wcoss_cray ]]; then
 
-  source $PATHTR/NEMS/src/conf/module-setup.sh.inc
```
junwang-noaa, line 102: My understanding is that this line actually does "module purge", which cleans up users' environment modules loaded before they run rt.sh. May I ask the reason we remove them?

DusanJovic-NOAA: I'm not sure about this line, maybe we can leave it, but fv3 module should not be loaded in the main script because ecflow server process remembers all the modules loaded before it's started.

junwang-noaa: Thanks, I think it is fine to remove loading the fv3 module  as we have separate module load at compile time and at run time.

PR #291, opened 2020-11-20
Port ufs-weather-model to acorn (WCOSS2 TDS)

### LINKED ISSUE #292: Port ufs-weather-model to acorn (WCOSS2 TDS)
## Description
We got access to acorn, a WCOSS2 TDS (Test Development System). UFS must be ported to WCOSS2.

## Solution
Add support to the build system and regression test to build and run tests on WCOSS2.

See PR #291

DusanJovic-NOAA:
## Description

Porting ufs-weather-model to acorn, WCOSS2 TDS. WCOSS2 is not using Intel CPU, so using `-xCORE-AVX2` compile flag creates executable that can not run on non-intel cpus. That flag is replaced by `-march=core-avx2`

### Issue(s) addressed
- fixes #292

## Testing

I ran one test (fv3_ccpp_control) manually on acorn. Input data for full regression tests are not yet available. This system has only 32 compute nodes, I'm not sure if we can run all tests using only 32 nodes.
Full test will be performed on all Tier-1 platforms. No changes in the baselines on Tier-1 platforms.

## Dependencies

N/A


tests/compile.sh
```diff
@@ -56,6 +56,9 @@ set +x
 if [[ $MACHINE_ID == macosx.* ]] || [[ $MACHINE_ID == linux.* ]]; then
   source $PATHTR/modulefiles/${MACHINE_ID}/fv3
 else
+  if [[ $MACHINE_ID == wcoss2 ]]; then
+    source /apps/prod/lmodules/startLmod
```
MinsukJi-NOAA, line 60: I guess this will eventually be moved to ../NEMS/src/conf/module-setup.*sh.inc?

modulefiles/wcoss2/fv3_debug
```diff
@@ -0,0 +1,42 @@
...
+
+module unload cpe-cray cce
+module load cpe-intel intel
+
+module load cmake/3.17.3
+setenv CMAKE_C_COMPILER cc
+setenv CMAKE_CXX_COMPILER CC
+setenv CMAKE_Fortran_COMPILER ftn
+setenv CMAKE_Platform wcoss2
+
```
aerorahul, line 17: One could just add `I_MPI_LINK` environment variable in the debug modulefile and eliminate the hard-wired compile time flags.

cmake/configure_wcoss2.cmake
```diff
@@ -0,0 +1,3 @@
+set(INLINE_POST     ON  CACHE BOOL "Enable inline post" FORCE)
+set(PARALLEL_NETCDF ON  CACHE BOOL "Enable parallel NetCDF" FORCE)
+set(DEBUG_LINKMPI   OFF CACHE BOOL "Enable linkmpi option when DEBUG mode is on" FORCE)
```
aerorahul, line 3: Are we debugging MPI implementation here?

I see we are hard-wiring flags such as:

```
if(DEBUG)
     if(OPENMP)
        set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -link_mpi=dbg_mt")
      else()
        set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -link_mpi=dbg")
      endif()
endif()
```

This should be handled by linking to the runtime debug MPI library by setting an environment variable such as `I_MPI_LINK`

By adding those flags at compile time, we are essentially debugging the MPI implementation.

aerorahul (commented):
Some comments on the unsavory use of MPI debug linkages

tests/rt.sh
```diff
@@ -173,6 +173,33 @@ elif [[ $MACHINE_ID = wcoss_dell_p3 ]]; then
   cp fv3_conf/fv3_bsub.IN_wcoss_dell_p3 fv3_conf/fv3_bsub.IN
   cp fv3_conf/compile_bsub.IN_wcoss_dell_p3 fv3_conf/compile_bsub.IN
 
+elif [[ $MACHINE_ID = wcoss2 ]]; then
+
+  source /apps/prod/lmodules/startLmod
```
MinsukJi-NOAA, line 178: Once NEMS/src/conf/module-setup.*sh.inc is modified to include wcoss2, this line will not be needed.

tests/compile.sh
DusanJovic-NOAA: Yes. Or if they switch to Lmod version of modules, sourcing this script might not be necessary at all.

DusanJovic-NOAA:
Merged in #295

Issue #285, opened 2020-11-18
Adding a benchmark test case for the fractional grid


ShanSunNOAA:
(1) Creating a fractional test "cpld_bmark_wave_frac" based on the existing "cpld_bmark_wave". 
(2) Additional initial conditions needed for this test will be added as well.

Issue #316, opened 2020-12-03
update restart tests for coupled model


DeniseWorthen:
## Description

After the merge of PR #304, we should update and/or modify existing restart tests for the coupled model. 

## Solution

This Issue has several aspects, some of which still need to be decided, so I consider this issue open for discussion. 

I believe that at a minimum, we should

- retire the c96mx025 restart test. This was carried over from ufs-s2s and should be replaced by a c96mx100 restart test if we want to retain a low resolution restart test.

- implement where possible checkpoint-restarting for the restart tests, reducing by one the number of tests that need to be run.

- implement a restart test for frac_grid.

- implement an 'overlap' restart test, meaning the the test will overlap the end of one day. Such a test would be for example a restart from hour 12, running for 36h and comparing to a continuous 48h forecast (12h/36h/48h). This references existing Issue #293. If this is implemented is a 'non-overlap' restart still required (12h/12h/24h)?

What is not clear yet to me is which resolutions should be tested for restart and how. 

- The benchmark+frac_grid configuration is the closest to what will be implemented, however it is also our most resource intensive test and we cannot include waves in a restart test at this point. Eventually this would also need to be a L127 test.

- If we have a benchmark+frac_grid restart test, are other restart tests (c96mx100) still required?

junwang-noaa:
I'd suggest to set up restart test with C96mx100 using the benchmarch+frac_grid configuration (except the resolution). If we have this setting I won't expect the restart will not work at high resolution benchmark+frac_grid test.

JessicaMeixner-NOAA:
For MOM6, the set-up is very different at 1deg versus 1/4 deg. Therefore, there are many aspects of the code that would be used operationally that would not be tested to really let us know about restart reproducibility in context of MOM6 if the restart test is only at 1deg.   I know there is a desire to make tests as small and short as possible, but this is likely not sufficient for MOM6 testing of restarts.  @jiandewang can provide more specific details if required.

junwang-noaa:
@JessicaMeixner-NOAA @jiandewang can you provide information on what are the features used in benchmark, but can not be used for low resolutions? Also would those features impact the coupled model in terms of model interface for coupled model? Can high resolution standalone MOM6 tests cover these feature testing including restart reproducibility? I am asking because ufs currently support 4 applications, so we do want to get fast RT turnaround time to avoid delays.

DeniseWorthen:
Currently, the plan is to:

[...]

PR #322, opened 2020-12-07
Adding a regression test "cpld_bmark_wave_frac"

ShanSunNOAA:
This is a benchmark regression test for C384L64, based on "cpld_bmark_wave". It uses a different set of input data, which needs to be copied over from /scratch2/BMC/gsd-fv3-dev/FV3-MOM6-CICE5/input-data-20201201_frac/FV3_input_frac/ to the default path, before the test is run. 

This refers to Issue #285.


DeniseWorthen:
After I get my bulk flux PR merged I was going to be opening a PR for updating the restart regression tests (Issue #316). I think it makes sense to combine this PR with the upcoming restart PRs. What do you think?

ShanSunNOAA:
I agree. The components are here to grep.

ShanSunNOAA:
This PR is replaced by #326, where the upper level codes are identical. The difference is the MOM6 submodule is uploaded correctly in #326.

DusanJovic-NOAA:
If this is replaced by #326 please close this PR.

ShanSunNOAA:
Just did. Sorry that I was unable to correct a submodule successfully in
#322, and decided to replace it with #326.

Yes, Denise, let us merge regression tests related into one PR.

Shan

PR #349, opened 2020-12-29
Update the format of rt.conf

### LINKED ISSUE #305: Need capability to exclude machines for tests in rt.conf
## Description
Currently we can either leave the list of machines in `rt.conf` empty to run a test on all machines, or we can specify a list of all machines to run on. Sometimes, tests need to be skipped on only one machine, which means that we have to repeat a `RUN` line several times.

## Solution
Implement a syntax that tells `rt.sh` to skip a certain test for the machines listed in the `MACHINES` column, for example by prepending the first entry with a `-`.

## Alternatives
Better solutions?

### LINKED ISSUE #329: rt.sh not running last test in rt*.conf when newline is missing at the end of the file
## Description

If the last test in `rt*conf` is not followed by a newline character, it is skipped by `rt.sh`.

### To Reproduce:
What compilers/machines are you seeing this with?
- hera.intel
Give explicit steps to reproduce the behavior.
1. Remove newline at end of `rt.conf` (or a smaller test version of it)
2. Run `rt.sh`
3. Check `log_hera.intel/` - the last test will be missed completely
4. then, oops, look at the bug!

DusanJovic-NOAA:
## Description

Update the format of `MACHINES` column in rt.conf (and other .conf files). This column can be either empty, which means a test will run on all supported machines, or start with `-` or `+` sign to exclude or include specified machines explicitly.

### Issue(s) addressed

- fixes #305
- fixes #329

## Testing

No baseline change.

## Dependencies

- waiting on https://github.com/NOAA-EMC/fv3atm/pull/218


climbfuji (commented):
This looks great, thanks for coming up with such a clean solution. I'll run the tests on gaea.intel, jet.intel, cheyenne.intel, cheyenne.gnu .

DeniseWorthen:
If I understand: 
if there are no specific machines listed, all machines will be tested. 
The (-) means that all machines except those listed as (-) will be tested, is that right? So it just removes those machines from the list of "all".
The (+) means that only those machines specifically listed will be tested.

tests/rt.conf
```diff
@@ -1,275 +1,198 @@
...
-RUN     | cpld_control                                                                                                 | standard | hera.intel orion.intel wcoss_dell_p3 | fv3 |
+COMPILE | DEBUG=Y                                                                                                                 | standard | + wcoss_dell_p3                         | fv3 |
+COMPILE | DEBUG=Y                                                                                                                 | standard | + wcoss_cray                            | fv3 |
+COMPILE | DEBUG=Y                                                                                                                 | standard | + hera.intel                            | fv3 |
+COMPILE | SUITES=FV3_GFS_v15p2,FV3_GFS_v16beta,FV3_GFS_v15p2_RRTMGP,FV3_GFS_v16beta_RRTMGP DEBUG=Y                                | standard | + orion.intel                           | fv3 |
+COMPILE | DEBUG=Y                                                                                                                 | standard | + cheyenne.intel                        | fv3 |
```
junwang-noaa, line 110: May I ask why the suites are specified on other platforms, but not on cheyenne? Are there other suite files compiled on cheyenne besides those listed on the line above?

tests/rt.conf
```diff
@@ -1,275 +1,198 @@
...
+RUN     | cpld_debug                                                                                                              | standard | + hera.intel orion.intel wcoss_dell_p3  | fv3 |
+#RUN     | cpld_debug                                                                                                             | standard | cheyenne.intel                          | fv3 |
+
+##############################################################################################################################################################################################
+# Data Atmosphere tests                                                                                                                                                                      #
+##############################################################################################################################################################################################
+
+COMPILE | DATM=Y S2S=Y                                                                                                            | standard | + hera.intel orion.intel wcoss_dell_p3  | fv3 |
```
junwang-noaa, line 179: I am curious, do we still allow platforms specified without "+" or "-", I feel the "+ platforms" is the same as "platforms", is it correct?

tests/rt.conf
climbfuji: The suites are only specified on orion.intel, but not on any of the other platforms (see lines 106-110). The reason is that compiling all suites on orion led to crash of the Intel compiler (some internal buffers overflowing or something like that).

tests/rt.conf
climbfuji: As far as I understand the code changes in `rt.sh`, a line can either be entirely empty (in which case it means all machines) or it **must** have a `+` or a `-` preceding a list of machine names. If there is no `+` or `-` and the line is not empty, this will create an error.

climbfuji:
Regression tests passed on cheyenne.intel, cheyenne.gnu, jet.intel, gaea.intel. All regression test logs are on hera and can be added to the PR (`/scratch1/BMC/gmtb/Dom.Heinzeller/ufs-weather-model/ufs-weather-model-emc-develop-20201229-cleanup-rt-et-al_logs`).

DusanJovic-NOAA:
> If I understand:
> if there are no specific machines listed, all machines will be tested.
> The (-) means that all machines except those listed as (-) will be tested, is that right? So it just removes those machines from the list of "all".
> The (+) means that only those machines specifically listed will be tested.

Correct.

tests/rt.conf
junwang-noaa: Do we know which suite files may cause issues on orion?

climbfuji: I don't think a particular suite causes an issue, it is the fact that there are "too many" of them and the compiler optimization algorithm can't handle it. Why only on orion and not on any of the other platforms, I don't know. But this happens, for example due to some internal buffer overflow.

DusanJovic-NOAA: I actually tried to compile the code on Orion with just "DEBUG=Y" and it worked. Compilation didn't crash. Maybe they updated compiler in the meantime or changed some other system configuration that was causing the crash before. I'm running a short test with this block in rt.conf

```
COMPILE | DEBUG=Y                            | standard | - gaea.intel jet.intel                  | fv3 |
RUN     | fv3_ccpp_gfs_v15p2_debug           | standard | + hera.intel orion.intel cheyenne.intel | fv3 |
RUN     | fv3_ccpp_gfs_v16beta_debug         | standard | - gaea.intel jet.intel                  | fv3 |
RUN     | fv3_ccpp_gfs_v15p2_RRTMGP_debug    | standard | + hera.intel orion.intel cheyenne.intel | fv3 |
RUN     | fv3_ccpp_gfs_v16beta_RRTMGP_debug  | standard | - gaea.intel jet.intel                  | fv3 |
```
If it passes we can make the change in one of the future commits.

DeniseWorthen (approved):

climbfuji (approved):
We should also do the following in the future: make corresponding changes to `tests/rt_ccpp_dev.conf` (I can do that when I pull the code into NOAA-GSL's gsl/develop branch, who maintain the file `tests/rt_ccpp_dev.conf`), and either remove `rt_stampede.conf` or make the corresponding changes there, too.

junwang-noaa (approved):

tests/rt.conf
DusanJovic-NOAA: These 4 tests passed on Orion.

climbfuji:
@DusanJovic-NOAA CI tests passed and you have three approvals, please merge whenever you are ready.

PR #369, opened 2021-01-13
dycore options to add zero-gradient BC to reconstruct interface u/v and change dz_min as input

XiaqiongZhou-NOAA:
## Description
only the dycore submodule in fv3atm is changed:
1) Add an option to use original BC or zero-gradient BC to reconstruct interface u/v with PSM for height advection
2)Change dz_min as a namelist input

### Issue(s) addressed
This is used to solve some model crashes in GFSv16 and regional forecasts.

## Testing
This is tested for GFSv16, regional forecast and idealized mountain wave runs.
This new BC option is not changing forecast performance and results significantly. 
It is not change regression test baseline since the default setting remains the same as the original

## Dependencies
https://github.com/NOAA-EMC/fv3atm/pull/232
https://github.com/NOAA-EMC/GFDL_atmos_cubed_sphere/pull/53


MinsukJi-NOAA (approved):
Changes related to CI and utest look good to me.

junwang-noaa (approved):

PR #382, opened 2021-01-17
Updated submodule for "Added new logical flag do_ugwp_v0_nst_only"

mdtoy:
Added new logical flag do_ugwp_v0_nst_only which allows non-stationary drag from ugwp_v0 to be run with GSL drag suite.

Note:  The same regression tests as before can be used.


climbfuji:
These changes were merged as part of https://github.com/ufs-community/ufs-weather-model/pull/396.

PR #397, opened 2021-01-27
release/public-v2: Make 32bit dynamics default (32BIT=ON) and set DYN32 accordingly

climbfuji:
## Description

The SRW App release team has decided that the default build for the ufs-weather-model should be with 32bit dynamics. 

Also, the CCPP switch that goes along with `32BIT=ON` for cmake should be set automatically in the cmake build, not in `compile_cmake.sh`. (A similar change was made for develop recently, but in a slightly different place due to the ongoing infrastructure updates in develop.)

Since all regression tests for release/public-v2 were run with `32BIT=Y`, no updates to the baselines are required.

## Issues

n/a

## Testing

Regression tests were run on all tier-1 platforms for release/public-v2, logs updated in the PR.

To make sure that compiling with 64bit dynamics still works, I also ran the following command on Jet:
```
./compile_cmake.sh $PWD/.. jet.intel 'CCPP=Y SUITES=FV3_GFS_v15p2,FV3_RRFS_v1alpha DEBUG=Y 32BIT=N' '' NO NO 2>&1 | tee compile_jet_intel_64bit.log
```
Compilation was successful and dynamics were compiled with double precision, see attached log: [compile_jet_intel_64bit.log](https://github.com/ufs-community/ufs-weather-model/files/5883920/compile_jet_intel_64bit.log)

## Dependencies

n/a

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ project(ufs-weather-model
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
-set(32BIT           OFF CACHE BOOL "Enable 32BIT (single precision arithmetic in dycore)")
+set(32BIT           ON  CACHE BOOL "Enable 32BIT (single precision arithmetic in dycore)")
 set(AVX2            ON  CACHE BOOL "Enable AVX2 instruction set")
 set(SIMDMULTIARCH   OFF CACHE BOOL "Enable multi-target SIMD instruction sets")
 set(CCPP            ON  CACHE BOOL "Enable CCPP")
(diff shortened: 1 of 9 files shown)


junwang-noaa (approved):
The changes look good. It looks to me all the RT tests are now use 32BIT, there is no 64bit test any more. I assume this release mainly supports 32BIT.

climbfuji:
> The changes look good. It looks to me all the RT tests are now use 32BIT, there is no 64bit test any more. I assume this release mainly supports 32BIT.

That's correct, the dynamics are 32bit for the release.

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ project(ufs-weather-model
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
-set(32BIT           OFF CACHE BOOL "Enable 32BIT (single precision arithmetic in dycore)")
+set(32BIT           ON  CACHE BOOL "Enable 32BIT (single precision arithmetic in dycore)")
```
christinaholtNOAA, line 20: @DomHeinzeller quick noob question....does the "CACHE" here mean that we could override this setting at a higher CMake level? Like if I specify this line (20) in the ufs-srweather-app/CMakeList.txt, will it take that setting when building the model? Would I maybe exclude "CACHE" when doing that to override?

climbfuji: I honestly don't know, but I think you are correct.

climbfuji:
@llpcarson @mkavulich do you want to take a look? I'd like to merge this before the code freeze tomorrow, if possible.

JeffBeck-NOAA (approved):

PR #399, opened 2021-01-29
changes for optional mesh use in MOM6; add dz_min and min_seaice as configurable variables for coupled model

### LINKED ISSUE #398: MOM6 cap should be switched using a Mesh at run time
The CMEPS mediator is mesh-based. Currently, the CICE6 component cap is also mesh based but the MOM6 component cap uses a grid.

For the purposes of mapping the ifrac/ofrac to the ATM, the source component is the ice component. This should be equivalent to taking those fields from the ocean component, since the ocean and ice run on the same grid and the fields are mapped by copying (with redist) from the ice to the ocean. However, testing has shown that when the MOM6 is on a grid and CICE6 is on the mesh, the ifrac and ofrac mapped to the Atm is different if the source is the ocean vs when it is the ice. 

This can be seen by temporarily inserting code in med_fraction_set.F90 which takes the ifrac and ofrac from the ocean component rather than the ice component. The following figures show the difference in ifrac (multipled by 1e14) mapped to the Atm when the compice is taken as the source component vs when the compocn is taken as the source component. In the C96 case, the differences are clearly aligned with the tripole seam. Differences also arise in the C192 case, though they are less striking.

The MOM6 cap can run on a mesh and when both MOM6 and CICE6 are run on a mesh, the the ifrac and ofrac can be mapped from either source component to the ATM with identical results.

To implement this in the MX100 case, MOM6 Issue [#47](https://github.com/NOAA-EMC/MOM6/issues/47) must first be resolved since the CMEPS code will abort when it identifies a mask point at run time which is different than the mesh file. 

Screen Shot 2021-01-28 at 7 08 43 PM

### LINKED ISSUE #411: add namelist variables for coupled model
## Description

Settings for dzmin and min_seaice should be added as configurable variables in the coupled tests.

## Solution

The default value of dz_min should be 2. For the L27 tests, the value should be set to 6. This is a change in ``fv_core_nml``.

The default value for min_seaice can be set to the current default value in the coupled model, which uses the puny value from CICE6 (1.0e-11). This is a change in ``gfs_physics_nml``.

DeniseWorthen:
## Description

*Implements an optional setting in the cpld and datm nems.configure files to specify whether the MOM6 cap should use a mesh or a grid. 

*Renames variable ``MESHICE`` to ``MESHOCN_ICE`` to clarify that MOM6 and CICE6 share a mesh file.

*Sets ``use_mommesh=false`` as a default to match the current use in ufs-weather.

*Adds configurable settings for  min_seaice to ``gfs_physics_nml`` and dz_min to ``fv_core_nml``. Default values are set to ``dz_min=2`` and ``min_seaice = 1.0e-11`` which match currently used values.

### Issue(s) addressed

- fixes #398 
- issue #411 

## Testing

Testing of the ufs-cpld and ufs-datm show no changes against current baselines since the mesh use is set false by default. Once MOM6 Issue [#47](https://github.com/NOAA-EMC/MOM6/issues/47) is resolved, a new mesh.mx100.nc can be generated and the default setting of ``use_mommesh`` can be switched to true.

Testing has also shown that the ufs-datm can be switched to the mesh for MOM6 w/o a change of baseline. In the datm application, no fields are being mapped back to the ATM component and all field between ocean and ice are copies (with redist). 

Switching to the mom mesh for ufs-cpld requires new new baselines.

In the future, ``dz_min`` can be set to 6 for v16, L127 layers tests. This is currently commented out in the benchmark v16 tests in order to retain current answers

## Dependencies

MOM6 PR [#54](https://github.com/NOAA-EMC/MOM6/pull/54)


climbfuji (approved):
I remember there was some discussion between you, Moorthi, Lucas and others as to the scientific reasoning for adjusting `dzmin`. I cannot say anything about this part, but from a code aspect the changes are fine.

junwang-noaa (approved):

DeniseWorthen:
Waiting on wcoss-dell. Still 9 jobs to go.

MinsukJi-NOAA (approved):

DeniseWorthen:
@MinsukJi-NOAA The CI tests currently are broken, is that correct?

junwang-noaa:
@DeniseWorthen CI tests are currently broken.

Issue #307, opened 2020-12-01
Use AWS EC2 instance for CI test


minsukji:
## Description
GitHub Actions-provided virtual machines are limited to in capacity, thus restricting different types of CI tests that can be run.

## Solution
Use AWS EC2 instance in the CI workflow.

PR #422, opened 2021-02-18
Use aws ec2 for CI test

MinsukJi-NOAA:
### Issue(s) addressed
#307 

## Testing

Changes in this PR are only related to CI (tests/ci directory) and Github Actions (.github/workflows directory), with only minor changes in default_vars.sh, run_test.sh, and utest.

The new CI tests were tested here: https://github.com/MinsukJi-NOAA/ufs-weather-model/actions/runs/579193301

## Dependencies

No dependencies.

### DIFF
--- a/tests/ci/build_status_check.py
+++ b/tests/ci/build_status_check.py
@@ -20,7 +20,6 @@ def update_url_data(response):
 
 def main():
 
-  time.sleep(600)
   url = json.load(sys.stdin)["workflow_run"]["jobs_url"]
 
   status="not-completed"
@@ -38,9 +37,9 @@ def main():
       status = "completed"
     else:
       no_completed_jobs = 0
-      time.sleep(10)
+      time.sleep(20)
 
-  time.sleep(10)
+  time.sleep(20)
   conclusion="failure"
   no_successful_jobs = 0
   for i in indices:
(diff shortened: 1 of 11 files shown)


tests/ci/build_status_check.py
```diff
@@ -0,0 +1,55 @@
...
+
+def update_url_data(response):
+  data = json.loads(response.read().decode())
+  indices=[]
+  for n in range(data["total_count"]):
+    if re.search("Build", data["jobs"][n]["name"]):
+      indices.append(n)
+
+  if len(indices) == 0:
+    sys.exit(1)
```
BrianCurtis-NOAA: `raise SystemExit('Some explanation here')` does the same thing, should try to raise an `Exception` instead of using sys.exit()

MinsukJi-NOAA: Done.

climbfuji (commented):
Do you want to leverage the existing S3 bucket at some point? I have been updating the input-data-YYYYMMDD in the bucket for the last several weeks, probably almost two months.

minsukji:
> Do you want to leverage the existing S3 bucket at some point? I have been updating the input-data-YYYYMMDD in the bucket for the last several weeks, probably almost two months.

I think it's definitely something that we need to look into at some point, especially as more test cases are added to CI.

Issue #427, opened 2021-02-22
WW3 not reproducing for different number of pets


JessicaMeixner-NOAA:
## Description
If you change the number of pets the WW3 model is using, the results do not reproduce the same answers as before.  For the wav-atm cases, just the restart file is different however, for the fully coupled case the output from the wave model, and the restarts are different.   This was discovered when trying to make a PR to put in the fix for the thread-safe issue, which is why that PR has not yet been submitted.  Details of the fixing this issue, since they are in WW3 will be in the WW3 issue that is related to this: https://github.com/NOAA-EMC/WW3/issues/317 

### To Reproduce:
Change the number of pets in a test with waves, the results will be different than before.

JessicaMeixner-NOAA:
So the fv3_ccpp_gfdlmprad_atmwav and fv3_ccpp_gfdlmprad tests reproduce themselves even with different number of MPI tasks, however, there are no restart checks in these tests. 

For the cpld_control_wave test, everything but the wave restart file is the same when using different number of MPI tasks for waves. 

For the cpld_bmarkfrac_wave and cpld_bmarkfrac_wave_v16 tests (and likely their non-frac counter parts), the tests does not pass when using different number of MPI tasks and even the atm files along with wave gridded output and restarts are different in this case.

JessicaMeixner-NOAA:
For the cpld_bmarkfrac_wave_v16 test here is an example of the magnitude of differences after 12 hours: 
<img width="566" alt="Screen Shot 2021-02-24 at 9 46 14 AM" src="https://user-images.githubusercontent.com/22549231/109017470-35300700-7685-11eb-960f-18dcf55fd8c0.png">
<img width="564" alt="Screen Shot 2021-02-24 at 9 45 22 AM" src="https://user-images.githubusercontent.com/22549231/109017480-36613400-7685-11eb-9791-b611f6905542.png">
<img width="563" alt="Screen Shot 2021-02-24 at 9 44 08 AM" src="https://user-images.githubusercontent.com/22549231/109017485-382af780-7685-11eb-85d1-8057204ae139.png">

junwang-noaa:
@JessicaMeixner-NOAA May I ask if this issue is resolved in WW3?

JessicaMeixner-NOAA:
Not yet, but @MatthewMasarik-NOAA has been looking into what might be causing this just from the WW3 standalone side and has an idea of what one issue there is.  Hopefully solving that problem will also solve this issue.

[...]

PR #383, opened 2021-01-18
Updatetemplate; Fix model slowness when using threading; Update WW3 for porting and threading issues

DeniseWorthen:
* Updates PR template with added PR checklist
* Fix model slowness when using threading [#258](https://github.com/NOAA-EMC/fv3atm/pull/258)
* Updates WW3 to the latest develop branch, which has updates for threading (see Issue #367). It also has updates for WW3 to be able to be build on gaea, jet and theoretically with gnu compilers although that has not been tested.

Dependent PRs:

FV3-atm: https://github.com/NOAA-EMC/fv3atm/pull/258

RT logs:

- [x] hera.intel
- [x] hera.gnu
- [ ] orion.intel
- [x] cheyenne.intel 
- [x] cheyenne.gnu
- [x] gaea.intel 
- [x] jet.intel
- [x] wcoss_cray
- [x] wcoss_dell_p3

PR #435, opened 2021-02-23
Update WW3 for porting and threading issues 

### LINKED ISSUE #367: Fully coupled test case (cpld_bmarkfrac_wave_v16) not thread-safe
## Description
It has been reported by @SMoorthi-emc and @shansun6 that when running cpld_bmarkfrac_wave_v16 test with and without threads do not reproduce each other.   

### To Reproduce:

Run cpld_bmarkfrac_wave_v16 and then run cpld_bmarkfrac_wave_v16 with multiple threads 

## Additional context

Going to try to run with different aspects of wave coupling turned on or off (ie, no wave feedback to ocean) to narrow down area to look for what part of the code is not thread safe.  Will update when have more information. 

@SMoorthi-emc to turn off OMP parts of WW3, remove [OMPG OMPH](https://github.com/NOAA-EMC/WW3/blob/develop/model/esmf/switch#L6-L7)  from WW3/model/esmf/switch

JessicaMeixner-NOAA:
## Description

This PR updates WW3 to the latest develop branch, which has updates for threading (see Issue #367)
it also has updates for WW3 to be able to be build on gaea, jet and theoretically with gnu compilers although that has not been tested.  

### Issue(s) addressed

- fixes #367

## Testing

How were these changes tested? On hera 
What compilers / HPCs was it tested with? intel 
Are the changes covered by regression tests? No, tests should be added for testing WW3 thread safe and tests should be turned on for waves on other platforms.   The WW3 thread safe test ran into an issue that we are having issues with the number of MPI tasks changing answers (Issue #427 ) which will be addressed in another PR.   The code compiles on Gaea but I did not have access to an account that can run the tests (request has been submitted), so the tests were not turned on.  Theoretically the code should be able to compile on jet now as well. 
Have regression tests and unit tests (utests) been run? On which platforms and with which compilers? hera.intel so far

## Dependencies

The PR in WW3 is merged https://github.com/NOAA-EMC/WW3/pull/315


BrianCurtis-NOAA:
Does this fix the bug that WW3 was crashing at runtime if built with -qopenmp?

JessicaMeixner-NOAA:
> Does this fix the bug that WW3 was crashing at runtime if built with -qopenmp?

@BrianCurtis-NOAA I don't actually remember that issue, can you help refresh my memory?  WW3 should be getting compiled with -qopenmp:  https://github.com/NOAA-EMC/WW3/blob/develop/model/bin/cmplr.env#L160-L167

BrianCurtis-NOAA:
I can't remember why but It was built OK with -qopenmp but somehow kept crashing at runtime.
If you want it to build with -qopenmp in UFS you will want to un-comment the multiple
`#if(OpenMP_Fortran_FOUND)
#  target_link_libraries(mom6 PRIVATE OpenMP::OpenMP_Fortran)
#endif()`
in MOM6-interface/CMakeLists.txt

JessicaMeixner-NOAA:
@BrianCurtis-NOAA WW3 is not yet getting options from the upper level cmake for build, so I'm puzzled to why other models using openMP would cause WW3 to crash? I know the atm-wav coupled can both use open mp.  Since you put MOM6 as your example, I'm wondering if this is an MOM6/S2S issue that you see even without WW3?   I don't think this PR would solve the issue you are describing.  This was just for if you ran WW3 with 1 or 2 threads that it would give you the same answer when two-way coupled to the atm model.

BrianCurtis-NOAA:
I see I read this wrong. You can ignore me.

JessicaMeixner-NOAA:
Merged with the latest ufs-weather-model and also updated WW3 to the top of it's develop branch.  hera.intel logs updated. 

This PR could be combined with another one, or whatever is most convenient.

DeniseWorthen:
Merged in PR #383

PR #443, opened 2021-02-26
Updates to auto-rt scripts to run on Cheyenne etc

climbfuji:
# WORK IN PROGRESS

## Description

Updates to auto-rt scripts to run on Cheyenne, temporary workaround to use GNU instead of Intel on that system. This will be generalized in a future PR.

### Issue(s) addressed

n/a

## Testing

# WORK IN PROGRESS

Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)
Have regression tests and unit tests (utests) been run? On which platforms and with which compilers? (Note that unit tests can only be run on tier-1 platforms)

## Dependencies

This PR will be pulled into and merged as part of https://github.com/ufs-community/ufs-weather-model/pull/425.

### DIFF
--- a/tests/auto/rt_auto.py
+++ b/tests/auto/rt_auto.py
@@ -42,7 +42,7 @@ def parse_args_in():
     parser = argparse.ArgumentParser()
 
     # Setup Input Arguments
-    choices = ['hera.intel', 'orion.intel', 'gaea.intel', 'jet.intel', 'wcoss_dell_p3']
+    choices = ['cheyenne.gnu', 'hera.intel', 'orion.intel', 'gaea.intel', 'jet.intel', 'wcoss_dell_p3']
     parser.add_argument('-m', '--machine', help='Machine and Compiler combination', required=True, choices=choices, type=str)
     parser.add_argument('-w', '--workdir', help='Working directory', required=True, type=str)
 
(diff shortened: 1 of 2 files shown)


climbfuji:
@BrianCurtis-NOAA Sorry for the back and forth, but given that this PR touches two files only, which are not used by anything else, it would be best to commit it separately afterwards. This way we can use the PR itself to demonstrate that the auto-rt updates are working as expected. Shouldn't be any more work.

DusanJovic-NOAA:
> @BrianCurtis-NOAA Sorry for the back and forth, but given that this PR touches two files only, which are not used by anything else, it would be best to commit it separately afterwards. This way we can use the PR itself to demonstrate that the auto-rt updates are working as expected. Shouldn't be any more work.

How about keeping these two files in a separate location (repository) during this early stages of development, to allow faster commits. Do they need to be located in the ufs-weather-model repository? If they are in some other location you will be able to make much more frequent changes. Just keep cron jobs pointing to the latest version.

BrianCurtis-NOAA:
@DusanJovic-NOAA I can manually update the scripts on each machine through development, but the final additions Jun would like to keep in the ufs-weather-model repo. 

@climbfuji I put the changes that were in this repo into PR#444. PR 444 only touches two files as well. The log files shouldn't get merged.

PR #473, opened 2021-03-12
Define applications but provide flexibility to build applications

### LINKED ISSUE #416: Having more flexible Cmake build system
## Description
I am currently working on integrating CDEPS data components with the model under HAFS application. This also requires extensive work related with the Cmake build system to build different data component configurations as well as adding PIO to support both CMEPS and CDEPS. The good news is that the initial version is ready in the following repository that supports DATM+DOCN and DATM+HYCOM configurations (FV3+DOCN will be available soon). 

https://github.com/hafs-community/ufs-weather-model.git - feature/hafs_couplehycom_cdeps branch

While I was working with the Cmake build I had couple of issues related with the Cmake build system and I would like to bring them to your attention. I think that these needs to be addressed in the near future.

- The current Cmake build system is not flexible enough to turn on/off individual model components and this requires lots of customization in the Cmake build system especially to support new data components provided by the CDEPS and other possible applications/components. One example for it is that FMS needs to build in all the cases even if I build CDEPS DATM+HYCOM case, that does not require FMS. Currently, there is no way to activate FMS only if FV3 or MOM6 are activated. Also, there is no way to turn on/off MOM6 and CICE explicitly such as MOM6=Y or CICE=Y. Those are just controlled by S2S build shortcut, which also makes the build more complicated. I think that those components also need to follow the same implementation way like WW3 and HYCOM. Since data components will be available soon and having multiple ocean components like MOM6, HYCOM, having flexible build system that allows to create different combinations of model components by turning on/off them easily by defining dependency between them will be crucial for the UFS model. This will simplify the top level Cmake file and also enable easy development and integration of other applications such as HAFS.

-  S2S shortcut is also creating another confusion because rather than tuning on CMPES, MOM6 and CICE explicitly. You could just set S2S=Y. This does not follow the current convention and requires additional customization when new dependencies and components will be available. One example of it is that i had problem related with supporting S2S related configurations by extending PIO support for CMEPS and CDEPS under the modeling system. It took extensive time for me to figure out that S2S also requires additional hack to make S2S RT test runs without any problem. The shortcuts like S2S can be still available but maybe the implementation logic can be revisited such as if you set S2S=Y it could trigger CMEPS=Y, MOM6=Y and CICE=Y rater then adding/activating them inside the S2S block or defining their dependencies such as FMS.

## Solution
The easiest solution is to follow the same logic used in WW3 and HYCOM and allow explicit control for all components including FV3, FMS etc. This will allow to user build only required components and reduce the executable size and improve the model performance slightly. 

## Alternatives
N/A

## Related to
N/A

aerorahul:
## Description

As the number of applications within the UFS grows, this PR collects the applications and re-defines it in terms of components that need to be enabled/disabled for the application.  It also allows the user the ability to define their own application based on a unique App Name.

The currently supported valid applications are:
|App Name| App Description| Components this App will turn ON|
|----------|-----------------|------------------|
| ATM    | Atmosphere Only | FMS, FV3 |
| ATMW | Atmosphere with Waves | FMS, FV3, WW3 |
| S2S    |  Coupled model | FMS, FV3, MOM6, CICE6, CMEPS |
| S2SW | Coupled model with Waves | FMS, FV3, MOM6, CICE6, WW3, CMEPS |
| DATM | Coupled Model with (CDEPS) Data atmosphere | FMS, MOM6, CICE6, CMEPS, CDEPS|
| DATM_NEMS | Coupled Model with (NEMS) Data atmosphere | FMS, MOM6, CICE6, CMEPS, NEMSdatm|

CMake expects an option `-DAPP=<APP_NAME>`, where `APP_NAME` is the item in the 1st column from the above table.
Without a valid application, the configuration will exit with a message.

To add new applications, the user will define a unique application name in  the list defined as `VALID_APPS` in `CMakeLists.txt` and proceed to enable the components for the new application in `cmake/configure_apps.cmake`.

What bug does it fix, or what feature does it add? 
Checks for valid applications as well as allows the user to define their own unique application with user choice of combination of components.

Is a change of answers expected from this PR?
No

Are any library updates included in this PR (modulefiles etc.)?
No

### Issue(s) addressed

Link the issues to be closed with this PR, whether in this repository, or in another repository.
(Remember, issues should always be created before starting work on a PR branch!)
Furthers the discussion in #416 
Possibly closes #416 

## Testing
On-going. Hera and Orion are very unresponsive!
Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)
Have regression tests and unit tests (utests) been run? On which platforms and with which compilers? (Note that unit tests can only be run on tier-1 platforms)

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
None.

Do PRs in upstream repositories need to be merged first?
No


DeniseWorthen:
The DATM is also a coupled model (MOM6-CICE6-CMEPS) using a DATM (NEMSdatm, replaced by CDEPS).

aerorahul:
> The DATM is also a coupled model (MOM6-CICE6-CMEPS) using a DATM (NEMSdatm, replaced by CDEPS). 

Thanks @DeniseWorthen I updated the App Description in the table and made it explicit. 

Do note that all these apps are in current develop.  I have not made an attempt to define what's coming or in various forks.

aerorahul:
I do not now why the indicated files changed in this PR is 204.
8 source code files have been changed in this branch as seen [here](https://github.com/ufs-community/ufs-weather-model/compare/develop...NOAA-EMC:feature/compsets)

aerorahul:
closing.  Merged via #477

Issue #644, opened 2021-06-15
move diag_table/field_table from input directory to tests/parm in the repository 


junwang-noaa:
## Description
Currently the diag_table and field_table are located in the input data directory. There are some discussion to move those small model test related configuration files to test/parm directory so that updates in those files can be better tracked. 
Another updates are to clean up the template variable replacement. Currently there are two different ways to replace the variable values in the template, it would be good if we can clean them up to avoid confusion.

## Solution
1. Move diag_table/field_table from input data directory to tests/parm directory and modify RT script to copy correct table in the run directory.
2. Use @[] in the templates for variable replacement.

## Related to
Directly reference any issues or PRs in this or other repositories that this is related to, and describe how they are related. Example:
- required to support noaa-emc/nems/issues/<issue_number>
- needed by noaa-emc/fv3atm/pull/<pr_number>

junwang-noaa:
@BinLiu-NOAA I noticed in your last commit, the field_table was changed back to the one in input directory (input-data-20210930/FV3_hafs_input_data/field_table), while in the PR [#870](https://github.com/ufs-community/ufs-weather-model/pull/870) the field_table is from ufs-weather-model/tests/parm/field_table/field_table_hafs. I am not removing all the field_tables in the input directory, but the HAFS tests will change due to the difference in the two field_tables. Do you want to change the ufs-weather-model/tests/parm/field_table/field_table_hafs to input-data-20210930/FV3_hafs_input_data/field_table?

BinLiu-NOAA:
@junwang-noaa, please help to delete the FV3_hafs_input_data/field_table file when preparing a new version INPUT DATA. We would like to use this one (ufs-weather-model/tests/parm/field_table/field_table_hafs) instead. BTW, please also help to delete this FV3_hafs_regional_input_data directory (which is no longer needed). Thanks!

junwang-noaa:
Bin, thanks for the clarification on the field_table.
The FV3_hafs_regional_input_data is deleted in the input
directory input-data-20211015 on hera, the input data directory will be
synced to other platforms when all the testing is done.

BinLiu-NOAA:
Thanks @junwang-noaa!

JessicaMeixner-NOAA:
@junwang-noaa we could remove all of the older WW3_input_data_* directories, except for the most recent one if you are working to clean-up and create a new input directory.

junwang-noaa:
[...]

PR #759, opened 2021-08-20
update templates for diag_table to be consistent with model_configure

aerorahul:
# PR Checklist

- [x] Ths PR is up-to-date with the top of all sub-component repositories except for those sub-components which are the subject of this PR.

- [ ] This PR has been tested using a branch which is up-to-date with the top of all sub-component repositories.  There is no change to the sub-components associated with this PR.

- [x] An Issue describing the work contained in this PR has been created either in the subcomponent(s) or in the ufs-weather-model.  See issue #644 

## Description

The `diag_table` templates (and other e.g. MOM6, WW3, ICE) employ different templating than the ones used in `input.nml`, `model_configure` and `nems.configure`.  These files use the template defined via `@[VARIABLE]` which is replaced by `$VARIABLE` defined upstream.

The workflow team would like to use the templates from the UFS-weather-model for constructing the necessary input files.
This will allow the workflow and the model to use the same templates, albeit filled with application specific values for the templated variables.

No change of answers is expected from this PR.
There is no impact on the general user running regression tests.
There are no library or modulefiles updated or need changes with this PR.

### Issue(s) addressed

- Part of #644 

## Testing

These have not been tested by running the model on any platform in particular with the exception that the change produces the desired output `diag_file` from the template as before.

- [ ] hera.intel
- [ ] hera.gnu
- [ ] orion.intel
- [ ] cheyenne.intel 
- [ ] cheyenne.gnu
- [ ] gaea.intel 
- [ ] jet.intel
- [ ] wcoss_cray
- [ ] wcoss_dell_p3
- [ ] CI

## Dependencies

None.

### DIFF
--- a/tests/edit_inputs.sh
+++ b/tests/edit_inputs.sh
@@ -52,11 +52,12 @@ function edit_data_table {
 }
 
 function edit_diag_table {
-  sed -e "s/YMD/$SYEAR$SMONTH$SDAY/g" \
-      -e "s/ATMRES/$ATMRES/g" \
-      -e "s/SYEAR/$SYEAR/g" \
-      -e "s/SMONTH/$SMONTH/g" \
-      -e "s/SDAY/$SDAY/g"
+  sed -e "s/@[YMD]/$SYEAR$SMONTH$SDAY/g" \
+      -e "s/@[ATMRES]/$ATMRES/g" \
+      -e "s/@[SYEAR]/$SYEAR/g" \
+      -e "s/@[SMONTH]/$SMONTH/g" \
+      -e "s/@[SDAY]/$SDAY/g" \
+      -e "s/@[SHOUR]/$SHOUR/g"
 }
 
 function edit_ww3_input {
--- a/tests/edit_inputs.sh
+++ b/tests/edit_inputs.sh
@@ -51,15 +51,6 @@ function edit_data_table {
   sed -e "s/FRUNOFF/$FRUNOFF/g"
 }
 
-function edit_diag_table {
-  sed -e "s/@[YMD]/$SYEAR$SMONTH$SDAY/g" \
-      -e "s/@[ATMRES]/$ATMRES/g" \
-      -e "s/@[SYEAR]/$SYEAR/g" \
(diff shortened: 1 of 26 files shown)


DeniseWorthen:
@aerorahul Just FYI, the "templating" files were initially introduced when S2S was being run using the Python based CompsetRunner that Sam built. They are overdue for updating to whatever works best now that all applications are using the rt.sh infrastructure.

tests/edit_inputs.sh
```diff
@@ -52,11 +52,12 @@ function edit_data_table {
...
-      -e "s/SYEAR/$SYEAR/g" \
-      -e "s/SMONTH/$SMONTH/g" \
-      -e "s/SDAY/$SDAY/g"
+  sed -e "s/@[YMD]/$SYEAR$SMONTH$SDAY/g" \
+      -e "s/@[ATMRES]/$ATMRES/g" \
+      -e "s/@[SYEAR]/$SYEAR/g" \
+      -e "s/@[SMONTH]/$SMONTH/g" \
+      -e "s/@[SDAY]/$SDAY/g" \
+      -e "s/@[SHOUR]/$SHOUR/g"
 }
```
DusanJovic-NOAA: This function is not necessary anymore. We should use atparse.

aerorahul: Would that mean replace:

```
edit_diag_table < ${PATHRT}/parm/${DIAG_TABLE:-diag_table_template} > diag_table
```
wtih
```
atparse < ${PATHRT}/parm/${DIAG_TABLE:-diag_table_template} > diag_table
```
on [L123](https://github.com/aerorahul/ufs-weather-model/blob/1e48750fb1d853c09c74616c1bfad086dea63e6e/tests/run_test.sh#L123) of `run_test.sh`?

junwang-noaa: That is correct. We are going to replace all the edit_ functions with atparse. Would you please make this change, then we can commit it? Thanks

aerorahul:
Github PR shows 29 files changed, when there are only 5 actually changed.
Closing and reopening this PR.

PR #1487, opened 2022-11-10
Remove loading anaconda module on hera

DusanJovic-NOAA:
This PR removes loading anaconda module on hera. It is not needed during build or model execution. This avoids conflict with different version of anaconda module loaded by GSI in global-workflow.


KateFriedman-NOAA (approved):
Tested this on Hera (build works) and tested resulting model exec at runtime (works). Thanks for making this change to resolve conflict with the GSI in the v16.3 system on Hera.

PR #1544, opened 2022-12-20
HAFSv1 needed HYCOM, UPP, g2tmpl, ugwp, and moving nesting related updates

BinLiu-NOAA:
# PR Checklist

- [x] This PR is up-to-date with the top of all sub-component repositories except for those sub-components which are the subject of this PR. Please consult the ufs-weather-model [wiki](https://github.com/ufs-community/ufs-weather-model/wiki/Making-code-changes-in-the-UFS-weather-model-and-its-subcomponents) if you are unsure how to do this.

- [x] This PR has been tested using a branch which is up-to-date with the top of all sub-component repositories except for those sub-components which are the subject of this PR

- [x] An Issue describing the work contained in this PR has been created either in the subcomponent(s) or in the ufs-weather-model. The Issue should be created in the repository that is most relevant to the changes in contained in the PR. The Issue and the dependent sub-component PR
are specified below.

- [x] Results for one or more of the regression tests change and the reasons for the changes are understood and explained below.

- [x] New or updated input data is required by this PR. If checked, please work with the code managers to update input data sets on all platforms.

## Instructions: All subsequent sections of text should be filled in as appropriate.

The information provided below allows the code managers to understand the changes relevant to this PR, whether those changes are in the ufs-weather-model repository or in a subcomponent repository. Ufs-weather-model code managers will use the information provided to add any applicable labels, assign reviewers and place it in the Commit Queue. Once the PR is in the Commit Queue, it is the PR owner's responsibility to keep the PR up-to-date with the develop branch of ufs-weather-model.

## Description

Provide a detailed description of what this PR does. What bug does it fix, or what feature does it add? Is a change of answers expected from this PR? Are any library updates included in this PR (modulefiles etc.)?

HAFSv1 needed updates and changes, including:
- Update submodule FV3/UPP to point its latest develop branch and update postxconfig-NT-hafs.txt.
- Switch to use g2tmpl/1.10.2.
- Switch to use the unified ugwp scheme for HAFS CCPP suites and regression tests.
- Update submodule HYCOM and enable debug and check all build for the HYCOM component (from @danrosen25 and @BijuThomas-NOAA).
- Fix the artificial gravity wave issue caused by HAFS moving nest leading edge crossing steep topography by enable using same coarse grid topography in the moving nest leading edge (from @wramstrom).
- Bug fix in calculation of moving-nest halo weights identified by @BijuThomas-NOAA in DDEBUG=ON builds (from @wramstrom).

Notes: 
[...]

Issue #1161, opened 2022-04-07
Updating options for ufs-weather-model compile settings


junwang-noaa:
## Description
In the current ufs-weather-model, three options/modes are available for the compile settings: PROD, REPRO and DEBUG. The PROD and DEBUG are currently tested in the ufs-weather-model RT. There are some [discussions](https://github.com/ufs-community/ufs-weather-model/discussions/1138) on the reproducibility of PROD and REPROD modes and updating the compile options to support UFS with operation's performance, reproducibility and accuracy requirements. Here is the final compile settings that we will use in ufs RT tests

1. remove the REPRO mode
2. replace the name PROD with RELEASE
3. update the flags for RELEASE to
a. Get rid of -qopt-prefetch=3 -no-prec-div -no-prec-sqrt options
b. replace -fp-model source with -fp-model consistent
 
## Solution
Implement the suggested compile option changes in ufs-weather-model.

## Related to
Directly reference any issues or PRs in this or other repositories that this is related to, and describe how they are related. Example:
- https://github.com/ufs-community/ufs-weather-model/discussions/1138
- https://github.com/ufs-community/ufs-weather-model/issues/1103

Issue #1103, opened 2022-03-10
Decomposition test failure


MinsukJi-NOAA:
## Description
`cpld_decomp_p8` fails with a different domain decomposition

### To Reproduce:
What compilers/machines are you seeing this with? Intel/Hera
Give explicit steps to reproduce the behavior.
1. Check out the latest ufs weather model (`9b6b740`)
2. `cd ufs-weather-model/tests; ./rt.sh -n cpld_decomp_p8`. This test will PASS
3. Modify `ufs-weather-model/tests/tests/cpld_decomp_p8`:
```
diff --git a/tests/tests/cpld_decomp_p8 b/tests/tests/cpld_decomp_p8
index cbf1b68f..bbbde45b 100644
--- a/tests/tests/cpld_decomp_p8
+++ b/tests/tests/cpld_decomp_p8
@@ -66,8 +66,10 @@ export RESTART_INTERVAL="${RESTART_N} -1"
 
 export TASKS=$TASKS_cpl_dcmp
 export TPN=$TPN_cpl_dcmp
-export INPES=$INPES_cpl_dcmp
-export JNPES=$JNPES_cpl_dcmp
+#export INPES=$INPES_cpl_dcmp
+#export JNPES=$JNPES_cpl_dcmp
+export INPES=8
+export JNPES=3
 export THRD=$THRD_cpl_dcmp
 export WRTTASK_PER_GROUP=$WPG_cpl_dcmp
```
4. Repeat step 2. This test will FAIL.

For a comparison,
5. Check out the previous commit `38aa634` of the ufs weather model
6. Repeat steps 2, 3, and 4 above. Both tests will PASS.

junwang-noaa:
@MinsukJi-NOAA Is this issue showing up in previous revisions?

MinsukJi-NOAA:
> @MinsukJi-NOAA Is this issue showing up in previous revisions?

The previous version `38aa634` does not have this issue.

junwang-noaa:
Since the physics suite is changed and there is no dycore update in 9b6b740, it might be related to physics. 

@yangfanglin @JessicaMeixner-NOAA @DeniseWorthen FYI.

yangfanglin:
Is there any uncoupled atmos decomposition test ? Is it working ?

JessicaMeixner-NOAA:
Yes, there is an uncoupled decomposition test.  I'm running it with he 8,3 combination now.

[...]

PR #1554, opened 2022-12-31
Faster, safer, compilation options targeted for HAFS operational upgrade

### LINKED ISSUE #1553: HAFS planned operationational optimization options
## Description
There should be a regression test for the optimization options that are planned for the HAFS operational upgrade so that we know the model will continue to build and run with those options.

## Solution
Add such a regressions test.

## Alternatives
Ignore the problem and hope nothing breaks.

## Related to
HAFS operational upgrade.

Requested by @BinLiu-NOAA @ZhanZhang-NOAA @BijuThomas-NOAA

SamuelTrahanNOAA:
## Description
Adds a set of faster, safer, optimization options planned for the HAFS operational upgrade whose code freeze is in January 2023. The FV3 uses the operational HRRR optimization options, without -ip, since that option doesn't speed up the model on WCOSS2, and it does slow down the compilation.  To enable this, use `-DFASTER=ON` when running the ufs-weather-model cmake.

Also, regardless of cmake options, the HYCOM no longer uses `-xSSE4.2` option that shouldn't be turned on in any WCOSS2 build.

One test, `hafs_regional_storm_following_1nest_atm_ocn_wav`, runs with an executable that uses the new compilation options. That means the results from that test will change. The default optimization options, and the actions of other `-Dwhatever` flags, are not changed at all. Therefore, *no other tests should change results.*

### FV3 Compilation Option Change

The new FV3 compilation options use value-safe optimization via `-fp-model precise`, rather than `-fp-model fast` like the default FV3 optimization options. They are far faster because just about every optimization is turned on, other than the ones that are turned off automatically due to not being value safe. Most importantly, `-fp-model consistent` is turned off. That option is often misunderstood; the only place it serves any real purpose is Jet, a heterogeneous cluster, where it is possible to get the same results across several machines. The cost of that is a massive performance penalty, even with -fp-model fast. Any other floating-point consistency safety already comes in -fp-model precise.

Oh, what was that? You don't see -fp-model fast in the current optimization options?

It is on by default. The Intel compiler has four categories of fp-model options. One of them is the value safety vs. speed trade-off: precise, fast, and fast2. If you don't specify any of those options, you get -fp-model fast. The -fp-model consistent is another category, and really serves no purpose if fp-model precise is enabled. (If you're wondering, the often used -fp-model strict turns on several options, including -fp-model precise.)

These options speed up the HAFS by 12% on WCOSS2.

### HYCOM Compilation Option Change

HYCOM uses AMD chips with the Intel compiler, and none of the Intel compiler settings do a great job of optimizing for AMD. The craype module already loads the best platform target settings. Specifying additional ones will slow down the model, or cause it to crash, if they have any effect at all.

This HYCOM compilation option change has only a small effect on the runtime since FV3 is the slow component in our current configuration.

### Top of commit queue on: TBD
There are no sub-component PRs.

### Input data additions/changes
- [x] No changes are expected to input data.
- [ ] ~~There will be new input data.~~
- [ ] ~~Input data will be updated.~~

### Anticipated changes to regression tests:
- [x] Changes are expected to the following tests: all tests that include "hafs" in their name.

The HYCOM compilation options have changed, so all tests that use HYCOM will have new results. That includes all tests with "hafs" in their name, but no other tests.

## Subcomponents involved:
- [ ] AQM
- [ ] CDEPS
- [ ] CICE
- [ ] CMEPS
- [ ] CMakeModules
- [ ] FV3
- [ ] GOCART
- [ ] HYCOM
- [ ] MOM6
- [ ] NOAHMP
- [ ] WW3
- [ ] stochastic_physics
- [x] none

## Commit Queue Checklist:

- [x] Link PR's from all sub-components involved
- [x] Confirm reviews completed in sub-component PR's
- [x] Add all appropriate labels to this PR.
- [x] Run full RT suite on either Hera/Cheyenne with both Intel/GNU compilers
- [x] Add list of any failed regression tests to "Anticipated changes to regression tests" section.

## Linked PR's and Issues:
Fixes #1553

## Testing Day Checklist:
- [x] This PR is up-to-date with the top of all sub-component repositories except for those sub-components which are the subject of this PR.
- [ ] ~~Move new/updated input data on RDHPCS Hera and propagate input data changes to all supported systems.~~

### Testing Log (for CM's):
- RDHPCS
  - Intel
    - [x] Hera
    - [x] Orion
    - [x] Jet
    - [x] Gaea
    - [x] Cheyenne
  - GNU
    - [x] Hera
    - [x] Cheyenne
- WCOSS2
  - [x] Dogwood/Cactus
  - [x] Acorn
- CI
  - [ ] Completed
- opnReqTest
  - [ ] N/A
  - [ ] Log attached to comment


SamuelTrahanNOAA:
It would be great if @BinLiu-NOAA @ZhanZhang-NOAA and @BijuThomas-NOAA could review this, as they're the HAFS developers that helped test these options on WCOSS2.

BrianCurtis-NOAA:
It looks like you ran RT with new baselines. Since it does not look like baselines are changed can you please run against current baselines `./rt.sh -e > rt.out 2>&1 &`. We would only expect the new RT to fail. You can attach that log to a comment instead of needing to push it to your repo if you'd like.

Also with any new RT, we need ORT to be run with `./opnReqTest -e -c thr,mpi,dcp,rst,bit,dbg,fhz -n hafs_regional_storm_following_1nest_atm_ocn_wav_supafast > ort.out 2>&1 &` it will reject some of the tests as we don't support all ORT with regional applications. So you can keep removing those until it starts officially. Please put that log into a comment in this PR.

SamuelTrahanNOAA:
> It looks like you ran RT with new baselines. Since it does not look like baselines are changed can you please run against current baselines ./rt.sh -e > rt.out 2>&1 &. We would only expect the new RT to fail. You can attach that log to a comment instead of needing to push it to your repo if you'd like.

You are incorrect. I ran against old baselines by rsyncing the BL_DATE baseline directory into my directory. Only the new test has a new directory.

BrianCurtis-NOAA:
> > It looks like you ran RT with new baselines. Since it does not look like baselines are changed can you please run against current baselines ./rt.sh -e > rt.out 2>&1 &. We would only expect the new RT to fail. You can attach that log to a comment instead of needing to push it to your repo if you'd like.
> 
> You are incorrect. I ran against old baselines by rsyncing the BL_DATE baseline directory into my directory. Only the new test has a new directory.

OK great. It didn't appear that way with your log, which confused me. Thanks.

BrianCurtis-NOAA:
@jkbk2004 @DeniseWorthen Please review code before commit queue assignment.

BrianCurtis-NOAA (approved):

jkbk2004:
So, performance impact is mostly expected on wcoss2, right?

SamuelTrahanNOAA:
> So, performance impact is mostly expected on wcoss2, right?

No. It'll be faster everywhere.

HYCOM-interface/CMakeLists.txt
```diff
@@ -9,7 +9,11 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
   set(CMAKE_Fortran_FLAGS_DEBUG "-O0 -fcheck=bounds -ffpe-trap=invalid,zero,overflow,underflow" )
   set(CMAKE_Fortran_LINK_FLAGS "")
 elseif(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
-  set(CMAKE_Fortran_FLAGS "-g -traceback -xSSE4.2 -mcmodel=small -r8")
+  if(SUPAFAST)
```
BinLiu-NOAA: @SamuelTrahanNOAA please make this change (removing -xSSE4.2) for all intel compiler build options, not only just under SUPAFAST option. In other words, I think the if (SUPAFAST) condition is not needed here. Thanks!

HYCOM-interface/CMakeLists.txt
```diff
@@ -24,7 +28,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "GNU")
   set(CMAKE_C_FLAGS_DEBUG "-O0")
   set(CMAKE_C_LINK_FLAGS "")
 elseif(CMAKE_C_COMPILER_ID MATCHES "Intel")
-  set(CMAKE_C_FLAGS "-g -traceback -xSSE4.2 -mcmodel=small")
+  if(SUPAFAST)
+    set(CMAKE_C_FLAGS "-g -traceback -mcmodel=small")
+  else()
+    set(CMAKE_C_FLAGS "-g -traceback -xSSE4.2 -mcmodel=small")
+  endif()
```
BinLiu-NOAA: Same here, please make this change (removing -xSSE4.2) when using Intel Compiler. Thanks!

cmake/Intel.cmake
```diff
@@ -17,10 +17,15 @@ if(DEBUG)
...
-    set(FAST "-fast-transcendentals")
+    if(SUPAFAST)
+      set(CMAKE_Fortran_FLAGS_RELEASE "-O3 -fp-model precise -assume buffered_stdout -fno-alias -align all -debug minimal -qoverride-limits -ftz")
+      set(CMAKE_C_FLAGS_RELEASE       "-O3 -fp-model precise -debug minimal -qoverride-limits -ftz")
+    else()
+      set(CMAKE_Fortran_FLAGS_RELEASE "-O2 -debug minimal -qoverride-limits")
+      set(CMAKE_Fortran_FLAGS_RELEASE "${CMAKE_Fortran_FLAGS_RELEASE} -fp-model consistent")
+      set(CMAKE_C_FLAGS_RELEASE       "-O2 -debug minimal")
+      set(FAST "-fast-transcendentals")
+    endif()
```
BinLiu-NOAA, line 28: @SamuelTrahanNOAA, @junwang-noaa, @DusanJovic-NOAA, For this SUPAFAST build option, I was wondering if we can/need rename it into PRODUCTION (or PROD). Currently, I believe we have DEBUG, RELEASE options. Another thing related, should we also test this SUPAFAST/PRODUCTION/PROD option for all ufs-weather-model regression tests? Especially to test if the decomp/restart tests can pass the regression tests.

tests/rt.conf
```diff
@@ -205,6 +205,9 @@ RUN     | hafs_regional_storm_following_1nest_atm_ocn
 RUN     | hafs_regional_storm_following_1nest_atm_ocn_wav                                                                         | - jet.intel                             | fv3 |
 RUN     | hafs_global_storm_following_1nest_atm                                                                                   | - jet.intel                             | fv3 |
 
+COMPILE | -DAPP=HAFSW -DMOVING_NEST=ON -DCCPP_SUITES=FV3_HAFS_v1_thompson_noahmp_nonsst,FV3_HAFS_v1_thompson_noahmp,FV3_HAFS_v1_thompson_nonsst,FV3_HAFS_v1_thompson,FV3_HAFS_v1_gfdlmp_tedmf_nonsst,FV3_HAFS_v1_gfdlmp_tedmf,FV3_HAFS_v1_thompson_tedmf_gfdlsf -D32BIT=ON -DSUPAFAST=ON |                | fv3 |
+RUN     | hafs_regional_storm_following_1nest_atm_ocn_wav_supafast                                                                | - jet.intel                             | fv3 |
+
```
BinLiu-NOAA, line 209: @SamuelTrahanNOAA, @junwang-noaa, I am not sure if we would prefer to add a new HAFS regression test (trying to avoid adding more regression tests if not absolutely needed.). Alternatively, we can just use the hafs_regional_storm_following_1nest_atm_ocn_wav RT but test it with the SUPAFAST/PRODUCTION/PROD build option instead.

SamuelTrahanNOAA:
Somehow, I forgot to do the FV3 PR.

tests/rt.conf
SamuelTrahanNOAA: Yes, I will make that change. It means this PR will be answer changing.

cmake/Intel.cmake
SamuelTrahanNOAA: I think it should replace "FAST" but not until it has been more widely tested.

HYCOM-interface/CMakeLists.txt
SamuelTrahanNOAA: Testing this now.

HYCOM-interface/CMakeLists.txt
SamuelTrahanNOAA: Testing this now.

cmake/Intel.cmake
SamuelTrahanNOAA: I would like comment from others about a suggested name for this option before I change it.

tests/rt.conf
SamuelTrahanNOAA: Testing this now.

SamuelTrahanNOAA: Perhaps we should run all of the HAFS regression tests with this build option?

SamuelTrahanNOAA:
Wasn't quite awake in my last post. There is no FV3 PR that I forgot to make. All of the build options are set at the ufs-weather-model level; nothing below that overrides them (I checked).

tests/rt.conf
BinLiu-NOAA: @SamuelTrahanNOAA, Thanks! Currently, we can just change to use this SUPAFAST/FAST/PRODUCTION/PROD option for this hafs_regional_storm_following_1nest_atm_ocn_wav RT. Later on, we can consider using it for more HAFS related RTs, especially after passing more rigorous tests.

junwang-noaa:
@SamuelTrahanNOAA Have your run the reproducibility tests with the fast compile options? Do we have run-to-run reproducibility? Will the results change with different threads? If I remember correctly, the hafs test does not reproduce with different MPI tasks or in restart runs yet.

cmake/Intel.cmake
BinLiu-NOAA: @SamuelTrahanNOAA, yeah, FAST is another good naming option. Would like to hear opinions from others (@junwang-noaa, @DusanJovic-NOAA, and other code managers) on these naming options: FAST/SUPAFAST/PRODUCTION/PROD, etc. Thanks!

SamuelTrahanNOAA:
> Have your run the reproducibility tests with the fast compile options?

I did, and so did other people.

> Do we have run-to-run reproducibility?

Yes.

> If I remember correctly, the hafs test does not reproduce with different MPI tasks 

No, it still doesn't. Several people tried several approaches to get this working, but we've run out of time for the implementation.

> or in restart runs yet

I don't even know how to set up a coupled wave-ocean-atmosphere restart test.

> Will the results change with different threads?

I don't think this has been tested with any optimization options (or lack thereof) because the main focus was MPI task decomposition. Someone else would have to do this test on Dogwood, since Cactus is unusable at the moment, and Acorn is too small.

cmake/Intel.cmake
SamuelTrahanNOAA: There is already a "FAST" option, so we would be replacing something that other groups might already be using. Naming it "PROD" or "PRODUCTION" wouldn't be great, since the utility of the optimization options are not limited to production.

tests/rt.conf
SamuelTrahanNOAA: My current test only uses it for hafs_regional_storm_following_1nest_atm_ocn_wav

SamuelTrahanNOAA:
The automatic github tests are failing due to an internal error:

> cat: /home/runner/id_file: No such file or directory

I manually checked, and my repo is up-to-date with develop.

SamuelTrahanNOAA:
> Have your run the reproducibility tests with the fast compile options?

@junwang-noaa Are you asking me to run all of the tests in rt.conf this way? I don't even know if they'll work. Some may fail to build, getting the dreaded "internal compiler error," which happens more often at higher optimization levels.

My only target for this was the two operational HAFS configurations, due to lack of time; the model took 50% more resources than it should, but now it runs in its window with a minute or three to spare. If people want to use this for other configurations, they'll have to test it first.

I'll try an "all suites" build with the SUPAFAST=ON and see what happens.

cmake/Intel.cmake
DusanJovic-NOAA: Some time ago we decided to support only two build options, 'Debug' and "Release'. No special options like FAST/PROD/PRODUCTION etc. If the compiler options added or changed by this PR are bit-by-bit reproducible then we should use them in all test ie. they should be enabled by default whenever 'Release' build type is selected. If they are not reproducible they shouldn't be used in any test/configuration. Similar for 'Debug' build type. Debug build type is used for debugging, Release for all other use cases (including production) and it should be as fast as possible. I do not see the need for any other build option.

DusanJovic-NOAA:
What's the meaning of 'SUPA' in SUPAFAST?

SamuelTrahanNOAA:
> What's the meaning of 'SUPA' in SUPAFAST?

It sounds better than "SUPERFAST"

SamuelTrahanNOAA:
> I'll try an "all suites" build with the SUPAFAST=ON and see what happens.

It's been 96 minutes, and the "all suites" build hasn't finished compiling. It has reached the caps, which already take a long time to compile on the default optimizations.

BinLiu-NOAA:
> > I'll try an "all suites" build with the SUPAFAST=ON and see what happens.
> 
> It's been 96 minutes, and the "all suites" build hasn't finished compiling. It has reached the caps, which already take a long time to compile on the default optimizations.

@SamuelTrahanNOAA, you probably can just use the SUPAFAST build options for the default RELEASE build and generate new baseline for all standard rt.conf RTs, than rerun to compare against the new generated baseline. If all the regression tests passed, then it might indicate it is safe to use these options as the default RELEASE build for all RTs. If not, then we can consider it only for one of the HAFS regression tests (if that's case, we maybe can use a temporary name like HAFSFAST, instead of SUPAFAST, build option for now?).

SamuelTrahanNOAA:
> @SamuelTrahanNOAA, you probably can just use the SUPAFAST build options for the default RELEASE build and generate new baseline for all standard rt.conf RTs, than rerun to compare against the new generated baseline. If all the regression tests passed, then it might indicate it is safe to use these options as the default RELEASE build for all RTs. If not, then we can consider it only for one of the HAFS regression tests (if that's case, we maybe can use a temporary name like HAFSFAST, instead of SUPAFAST, build option for now?).

I did this, and all tests that were submitted passed: 102 tests and 25 compile jobs. Three compile jobs, for FV3_HRRR, were unable to finish compiling `ccpp_FV3_HRRR_cap.F90` with the high optimization levels, even with a 3hr wallclock limit. If we had a way to lower optimization levels on certain files, that would fix it. The operational HRRR had to use `-O1` (optimize for size) on the WRF framework files, so this isn't a surprise.

SamuelTrahanNOAA:
> Have your run the reproducibility tests with the fast compile options? Do we have run-to-run reproducibility? Will the results change with different threads? If I remember correctly, the hafs test does not reproduce with different MPI tasks or in restart runs yet.

@junwang-noaa The answer is: yes, things are reproducible. I was able to succeed 102 tests and 25 compile jobs on Hera. The FV3_HRRR suite couldn't finish compiling in time, so those tests were untested. We need to be able to reduce optimization on CCPP cap files for that suite to compile.

cmake/Intel.cmake
SamuelTrahanNOAA: You need `-fp-model consistent` on heterogenous clusters like Jet if you want the different partitions get the same result. You can't afford that luxury in operations. That means we do need three options:

1. Debugging. (The current "Debug" target.)
2. Support for heterogenous clusters. (The current "Release" target.)
3. Runs at a usable speed. (What this PR calls "SUPAFAST.")

SamuelTrahanNOAA:
I resubmitted the FV3_HRRR compile jobs with an 8hr wallclock limit. We'll see if they compile this time.

cmake/Intel.cmake
DusanJovic-NOAA: We always run tests on the same Jet cluster (xjet), if -fp-model precise is reproducible (is it?), then I do not think we need two different 'Release' versions, we should just use one.

SamuelTrahanNOAA: Unfortunately, the FV3_HRRR suite cap cannot compile in 8 hours. Until we have a way to set a different optimization option for the caps, this can't be used for all suites.

Does anyone know why `-fp-model consistent` was in the build options to begin with? Surely someone put it there for a reason?

SamuelTrahanNOAA:
> I resubmitted the FV3_HRRR compile jobs with an 8hr wallclock limit. We'll see if they compile this time.

Eight hours wasn't enough. I need a way to use different optimization options for the CCPP caps, and I don't see one in the cmake files.

junwang-noaa:
@SamuelTrahanNOAA Thanks for the testing! Somehow I have an impression that O3 does not guarantee run to run reproducibility and we had some weird output field with it, but that was long time ago.

SamuelTrahanNOAA:
> @SamuelTrahanNOAA Thanks for the testing! Somehow I have an impression that O3 does not guarantee run to run reproducibility and we had some weird output field with it, but that was long time ago.

I wouldn't use -O3 or other aggressive optimization options without -fp-model precise. The -O3 option does not guarantee reproducibility unless you select an -fp-model that disables non-value-safe optimizations (-fp-model precise) or non-reproducing optimizations (-fp-model consistent). It can lose a lot of precision in calculations unless you tell it not to (-fp-model precise). That certainly could result in strange values in the fields.

cmake/Intel.cmake
DusanJovic-NOAA: @climbfuji Do you remember when and why we started to use -fp-model consistent?

climbfuji:
Yes, that was b4b reproducibility way back, I think when the REPRO mode was still in use. There were one or two physics schemes (mg?) that needed “consistent” in order to be b4b reproducible. “precise” should be fine, as far as I remember it is more strict than consistent.

junwang-noaa:
@DusanJovic-NOAA I think it's because of GOCART tests. Please see issue [#1161](https://github.com/ufs-community/ufs-weather-model/issues/1161) and [#1103](https://github.com/ufs-community/ufs-weather-model/issues/1103).

SamuelTrahanNOAA:
Everyone: Incorrect report on my part. I only ran the -c, to create the baseline. I haven't run -m yet, so I don't know if it reproduces.

I'll run that now. Of course, the FV3_HRRR will fail since the baselines couldn't be generated.

BinLiu-NOAA:
@SamuelTrahanNOAA, Thanks for these additional tests and updates! Also, good to see that it at least can generate baseline for all non FV3_HRRR related tests. Looking forward to seeing whether or not the new regression tests can reproduce the new generated baseline. 

Regarding the compiler out of wall-clock issue for those FV3_HRRR regression tests, @junwang-noaa, @DusanJovic-NOAA and others, any suggestions or comments on how to deal with this situation? Maybe considering three-tier build options (DEBUG, RELEASE, FAST) might be an approach here. Different subsets of the regression tests can choose which build option they would like to use/test with. 

Another approach is probably as Sam mentioned to always use lower build options for those FV3_HRRR related source code files (which take too long when using -O3 build options). 

Alternatively, if I recall correctly, separating the related very long source code files into multiple short source code files can help to reduce the compile time. Not sure if this might help though.

Thanks!

junwang-noaa:
@SamuelTrahanNOAA @BinLiu-NOAA @RatkoVasic-NOAA  @ChunxiZhang-NOAA @yangfanglin  I am not sure if any ufs wm application is using fv3_hrrr for implementation, I saw those global hrrr_* tests, not sure who is using it and for what. If I remember correctly, these tests were set up by GSL, Does anyone know if anyone can maintain those tests? Whom should code managers to contact if there is issue with those tests?

SamuelTrahanNOAA:
Everyone: good news: all tests passed `-m` except the ones that tried to compile the FV3_HRRR suite.

DusanJovic-NOAA:
> Everyone: good news: all tests passed `-m` except the ones that tried to compile the FV3_HRRR suite.

Great. How much faster are tests running, for example control_p8, or cpld_control_p8?

SamuelTrahanNOAA:
> Great. How much faster are tests running, for example control_p8, or cpld_control_p8?

I don't know the typical runtime. Also, some of the job time is taken up by copying files before running srun. You'd need to run the UFS directly, with a longer forecast length, to get meaningful statistics.

Here's the runtimes for p8 though. The last column has the total job times in seconds.

```
197001010000    cpld_control_p8_mixedmode                    40875858           SUCCEEDED                   0         1         333.0
197001010000         cpld_control_p8                    40875883           SUCCEEDED                   0         1         362.0
197001010000    cpld_control_ciceC_p8                    40875889           SUCCEEDED                   0         1         370.0
197001010000    cpld_control_c192_p8                    40875888           SUCCEEDED                   0         1         618.0
197001010000    cpld_control_noaero_p8                    40875788           SUCCEEDED                   0         1         285.0
197001010000    cpld_control_nowave_noaero_p8                    40875789           SUCCEEDED                   0         1         284.0
197001010000    cpld_control_noaero_p8_agrid                    40875790           SUCCEEDED                   0         1         298.0
197001010000              control_p8                    40875771           SUCCEEDED                   0         1         192.0
197001010000         control_p8_lndp                    40875772           SUCCEEDED                   0         1         315.0
197001010000      control_restart_p8                    40875931           SUCCEEDED                   0         1         121.0
197001010000       control_decomp_p8                    40875773           SUCCEEDED                   0         1         191.0
197001010000     control_2threads_p8                    40875774           SUCCEEDED                   0         1         185.0
197001010000       control_p8_rrtmgp                    40875775           SUCCEEDED                   0         1         225.0
197001010000        control_debug_p8                    40875959           SUCCEEDED                   0         1         204.0
197001010000    control_p8_atmlnd_sbs                    40876148           SUCCEEDED                   0         1         228.0
197001010000      atmaero_control_p8                    40876175           SUCCEEDED                   0         1         251.0
197001010000    atmaero_control_p8_rad                    40876176           SUCCEEDED                   0         1         297.0
197001010000    atmaero_control_p8_rad_micro                    40876177           SUCCEEDED                   0         1         283.0
```

DusanJovic-NOAA:
> > Great. How much faster are tests running, for example control_p8, or cpld_control_p8?
> 
> I don't know the typical runtime. Also, some of the job time is taken up by copying files before running srun. You'd need to run the UFS directly, with a longer forecast length, to get meaningful statistics.
> 
> Here's the runtimes for p8 though. The last column has the total job times in seconds.
> 
> ```
> 197001010000    cpld_control_p8_mixedmode                    40875858           SUCCEEDED                   0         1         333.0
> 197001010000         cpld_control_p8                    40875883           SUCCEEDED                   0         1         362.0
> 197001010000    cpld_control_ciceC_p8                    40875889           SUCCEEDED                   0         1         370.0
> 197001010000    cpld_control_c192_p8                    40875888           SUCCEEDED                   0         1         618.0
> 197001010000    cpld_control_noaero_p8                    40875788           SUCCEEDED                   0         1         285.0
> 197001010000    cpld_control_nowave_noaero_p8                    40875789           SUCCEEDED                   0         1         284.0
> 197001010000    cpld_control_noaero_p8_agrid                    40875790           SUCCEEDED                   0         1         298.0
> 197001010000              control_p8                    40875771           SUCCEEDED                   0         1         192.0
> 197001010000         control_p8_lndp                    40875772           SUCCEEDED                   0         1         315.0
> 197001010000      control_restart_p8                    40875931           SUCCEEDED                   0         1         121.0
> 197001010000       control_decomp_p8                    40875773           SUCCEEDED                   0         1         191.0
> 197001010000     control_2threads_p8                    40875774           SUCCEEDED                   0         1         185.0
> 197001010000       control_p8_rrtmgp                    40875775           SUCCEEDED                   0         1         225.0
> 197001010000        control_debug_p8                    40875959           SUCCEEDED                   0         1         204.0
> 197001010000    control_p8_atmlnd_sbs                    40876148           SUCCEEDED                   0         1         228.0
> 197001010000      atmaero_control_p8                    40876175           SUCCEEDED                   0         1         251.0
> 197001010000    atmaero_control_p8_rad                    40876176           SUCCEEDED                   0         1         297.0
> 197001010000    atmaero_control_p8_rad_micro                    40876177           SUCCEEDED                   0         1         283.0
> ```

In the regression test log file we save the 'wallclock time' which is only model execution time.

SamuelTrahanNOAA:
> In the regression test log file we save the 'wallclock time' which is only model execution time.

I won't have those reports because the rt.sh will never finish due to FV3_HRRR.

SamuelTrahanNOAA:
> I won't have those reports because the rt.sh will never finish due to FV3_HRRR.

I've disabled the tests I expect to fail, and I'm running again to make sure all the rest succeed. This should also give us the timings in Regression_hera.intel.log.

DusanJovic-NOAA:
Try to use `-no-ip` flag, hopefully if will not have huge impact on execution speed but it should speed up the compilation.

SamuelTrahanNOAA:
I've tried disabling options to the caps, and it gets past those now, but eventually it gets stuck on cu_gf_deep.F90.

SamuelTrahanNOAA:
I don't think I can get all suites to compile in the near future due to my other operational upgrade responsibilities. (HAFS is not my only impending operational code freeze.) Is there some compromise we can come to? Maybe this?

1. Call it something like `-DEXPERIMENTAL_OPTIMIZATIONS=ON`.
2. Someone gets it working for all suites.
3. It is scientifically validated.
4. Then, it replaces the Release target's compilation options.

DusanJovic-NOAA:
On Hera, using -no-ip flag, I successfully created new baseline and verified against it. 

[RegressionTests_hera.intel.log](https://github.com/ufs-community/ufs-weather-model/files/10347868/RegressionTests_hera.intel.log)

SamuelTrahanNOAA:
> On Hera, using -no-ip flag, I successfully created new baseline and verified against it.

For us to use -no-ip, Biju would have to test the model on WCOSS2 with that option to see if it still fits in the operational window.

BinLiu-NOAA:
> > On Hera, using -no-ip flag, I successfully created new baseline and verified against it.
> 
> For us to use -no-ip, Biju would have to test the model on WCOSS2 with that option to see if it still fits in the operational window.

Thanks, @DusanJovic-NOAA and @SamuelTrahanNOAA! As for this -no-ip build option, I assume it should not affect the runtime substantially, correct? Dusan, could you point us to your exact changes for this -no-ip flag, we can ask @BijuThomas-NOAA's help to conduct two tests on WCOSS2 comparing with and without this -no-ip flag to see if there are noticeable differences for the HAFSv1 configuration's 126-h forecast wallclock time. Thanks!

SamuelTrahanNOAA:
Dusan's change could hurt performance on some physics schemes, but it can help performance on some more complicated code like the ccpp framework. There's no way to know what will happen until you try.

BijuThomas-NOAA:
> > > On Hera, using -no-ip flag, I successfully created new baseline and verified against it.
> > 
> > 
> > For us to use -no-ip, Biju would have to test the model on WCOSS2 with that option to see if it still fits in the operational window.
> 
> Thanks, @DusanJovic-NOAA and @SamuelTrahanNOAA! As for this -no-ip build option, I assume it should not affect the runtime substantially, correct? Dusan, could you point us to your exact changes for this -no-ip flag, we can ask @BijuThomas-NOAA's help to conduct two tests on WCOSS2 comparing with and without this -no-ip flag to see if there are noticeable differences for the HAFSv1 configuration's 126-h forecast wallclock time. Thanks!

I did a 126-h run on Dogwood with the build option -no-ip(in Intel.cmake)

``if(SUPAFAST)      
      set(CMAKE_Fortran_FLAGS_RELEASE "-O3 -fp-model precise -assume buffered_stdout -fno-alias -align all -debug minimal -qoverride-limits -ftz -no-ip")
      set(CMAKE_C_FLAGS_RELEASE       "-O3 -fp-model precise -debug minimal -qoverride-limits -ftz -no-ip")``

The run was completed with the total amount of wall time 5658.711410 (94mins)

SamuelTrahanNOAA:
> The run was completed with the total amount of wall time 5658.711410 (94mins)

Is that the version with WW3 or without?

I recall these numbers:

- FV3-HYCOM-WW3 was only 93 minutes (due to cheaper physics)
- FV3-HYCOM was 95.

So, if 93 became 94, we're seeing it slow down the model, but if 95 became 94, it is speeding up the model.

BinLiu-NOAA:
> The run was completed with the total amount of wall time 5658.711410 (94mins)
@BijuThomas-NOAA, thanks for testing this out quickly! If I recall correctly, this 94 min wall clock time is very similar  (within 1-2 mins which might can be considered as system side run-to-run variations) to the one without using this -no-ip build flag. If this is the case, @SamuelTrahanNOAA, @DusanJovic-NOAA, and @junwang-noaa, I think we can just add this -no-ip on top of what Sam has for SUPAFAST. Then the other question is, do we just want to use this set of SUPAFAST build options to replace the RELEASE build options, or keep the existing RELEASE build option while adding a new (saying FAST/PRODUCTION) build option? I can see both have pros and cons though. Thanks!

SamuelTrahanNOAA:
If it replaces the Release base, it'll affect all models (GFS, RRFS, etc.) and you don't know what impact it'll have on the forecast skill. These options need to be scientifically validated before they become the default. I think we should have an option to turn it on, and let everyone know it will be the default in a few weeks (or however long is appropriate) unless we get negative results.

BijuThomas-NOAA:
> > The run was completed with the total amount of wall time 5658.711410 (94mins)
> 
> Is that the version with WW3 or without?
> 
> I recall these numbers:
> 
> * FV3-HYCOM-WW3 was only 93 minutes (due to cheaper physics)
> * FV3-HYCOM was 95.
> 
> So, if 93 became 94, we're seeing it slow down the model, but if 95 became 94, it is speeding up the model.

@SamuelTrahanNOAA  This is the HAFS-B configuration without WW3(The test case I got from you).

SamuelTrahanNOAA:
> @SamuelTrahanNOAA This is the HAFS-B configuration without WW3(The test case I got from you).

Excellent. That means Dusan's change has sped up the model. I don't recall ever seeing a number lower than 5720 in my tests of that configuration.

junwang-noaa:
I agree with Sam. I'd suggest having a new option first to let developers test. Once it is confirmed that the new option improves computational performance and has compatible scientific results, we can then replace the RELEASE mode.  Do we have any scientific evaluation of HAFS results with the new compile options yet?

BinLiu-NOAA:
> If it replaces the Release base, it'll affect all models (GFS, RRFS, etc.) and you don't know what impact it'll have on the forecast skill. These options need to be scientifically validated before they become the default. I think we should have an option to turn it on, and let everyone know it will be the default in a few weeks (or however long is appropriate) unless we get negative results.

Agree with @SamuelTrahanNOAA! It might be good to stay on the conservative side currently, keeping the existing DEBUG and RELEASE build options while adding a FAST/PRODUCTION/whatever build option at this point. Later on, if absolutely needed, the RELEASE and FAST/PRODUCTION options can be merged back into one RELEASE option after comprehensive technical/scientific testing. Thanks!

SamuelTrahanNOAA:
There's already a "-DFAST=ON" which someone might be using, so we need a different name.

BinLiu-NOAA:
> I agree with Sam. I'd suggest having a new option first to let developers test. Once it is confirmed that the new option improves computational performance and has compatible scientific results, we can then replace the RELEASE mode. Do we have any scientific evaluation of HAFS results with the new compile options yet?

@junwang-noaa, we have not conducted comprehensive/scientific testing/evaluation for these SUPAFAST options from the HAFS side yet. But, that will be upcoming in the next few weeks for HAFSv1 final configurations implementation tests (with thousands of forecast cycles). We will definitely notice if these build options have substantial impacts (degradation) on forecast skills. Thanks!

junwang-noaa:
Thanks, Bin. @SamuelTrahanNOAA I'd suggest using "SUPERFAST" rather than "SUPAFAST" to avoid confusion.

BrianCurtis-NOAA:
I would suggest "TESTFAST" so people know we are testing with it.

SamuelTrahanNOAA:
I have called it `TEST_FASTER_INTEL` and I'm testing it now.

SamuelTrahanNOAA:
@DusanJovic-NOAA - could you please check my cmake/Intel.cmake to make sure I added the -no-ip correctly?

DusanJovic-NOAA:
> @DusanJovic-NOAA - could you please check my cmake/Intel.cmake to make sure I added the -no-ip correctly?

Looks good. You'll also need to update rt.conf with new -D option. If that new test config file is not needed anymore please delete it.

BinLiu-NOAA:
> I have called it `TEST_FASTER_INTEL` and I'm testing it now.

@SamuelTrahanNOAA, if possible and no other objections, could you please rename TEST_FASTER_INTEL into FASTER (preferred) or FASTER_INTEL. The reason is that we most likely will use this build option for HAFSv1 operational implementation, it would be better to avoid using something like "TEST_*" build option for HAFSv1 code delivery. Thanks!

SamuelTrahanNOAA:
> @SamuelTrahanNOAA, if possible and no other objections, could you please rename TEST_FASTER_INTEL into FASTER (preferred) or FASTER_INTEL. The reason is that we most likely will use this build option for HAFSv1 operational implementation, it would be better to avoid using something like "TEST_*" build option for HAFSv1 code delivery. Thanks!

Oh! Good point. Calling it a "TEST" would confuse NCO, wouldn't it. 

Does anyone object to using `-DFASTER=ON` until this becomes the default for Release?

SamuelTrahanNOAA:
@BinLiu-NOAA If I remove `-xSSE4.2` from all HYCOM builds, that'll change the results for a lot of tests, won't it?

SamuelTrahanNOAA:
I think it's all of the tests with "HAFS" in the app name.

BinLiu-NOAA:
> I think it's all of the tests with "HAFS" in the app name.
@SamuelTrahanNOAA, yes, that's correct. It will change all HAFS coupled RTs possibly.

SamuelTrahanNOAA:
>  It will change all HAFS coupled RTs possibly.

I'm regenerating those baselines now.

BinLiu-NOAA:
@jkbk2004, and other code managers, it looks like we have agreed with a way moving forward. Also, I believe this PR is up to date on top of the latest develop branch. Considering this PR is needed by PR #1544, which required by HAFSv1 code freeze, is it possible we can expedite the review, test and commit process for this PR? Much appreciated!

BinLiu-NOAA (approved):
All my review comments/concerns have been addressed.

SamuelTrahanNOAA:
It is ready for testing. I tested the previous version though; develop was updated since my last test, and I merged those changes in. I didn't see anything that could break the `-DFASTER=ON` though.

Definitely, it is time for someone to independently verify that this works, so I'd like the code managers to regenerate the hafs baselines and run the regression tests, on their platform of choice.

Also, I think I've resolved the problems Dusan found, but if he could re-review and confirm, that would be great.

SamuelTrahanNOAA:
I was able to generate baselines and validate against them using `-DFASTER=ON`. This is with the version before I merged develop; the same one I used to run the hera baselines.

dkokron:
I was able to reproduce the excessive compile time issue (without -no-ip).  We can avoid using the -no-ip flag on the whole code using the following patch to FV3/ccpp/physics/CMakeLists.txt

```
git diff CMakeLists.txt
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 24227541..52f4bb00 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,6 +178,12 @@ if(${LOCAL_CURRENT_SOURCE_DIR}/physics/rte-rrtmgp/rrtmgp/kernels/mo_gas_optics_k
                               APPEND_STRING PROPERTY COMPILE_FLAGS " ${CMAKE_Fortran_FLAGS_PHYSICS} -O1")
 endif()
 
+# Reduce optimization for cu_gf_deep.F90 to avoid excessive compile time at -O3
+if(CMAKE_BUILD_TYPE STREQUAL "Release" AND ${CMAKE_Fortran_COMPILER_ID} STREQUAL "Intel")
+  SET_SOURCE_FILES_PROPERTIES(${LOCAL_CURRENT_SOURCE_DIR}/physics/cu_gf_deep.F90
+                              APPEND_STRING PROPERTY COMPILE_FLAGS " ${CMAKE_Fortran_FLAGS_PHYSICS} -O2")
+endif()
+
 #------------------------------------------------------------------------------
 
 add_library(ccpp_physics STATIC ${SCHEMES} ${SCHEMES_OPENMP_OFF} ${SCHEMES_DYNAMICS} ${CAPS})
```

SamuelTrahanNOAA:
You may not be aware, but @dkokron is one of the analysts working on speeding up the HAFS. I'd like to test his fix before we call this optimization the final one. That may take several days, so if you want to go ahead with the PR as is, I can work with @dkokron to make a second PR with further optimization improvements.

Also, I suspect we can keep some of the speed improvements from -no-ip by using -O1 -no-ip or -O2 for the CCPP caps. I have some code on disk that'll do that (though, I suspect most of you could put it together easily).

BinLiu-NOAA:
@SamuelTrahanNOAA and @jkbk2004, since we are waiting this PR #1554 and the upcoming PR #1544 for HAFSv1 code freeze (targeting before end of next week 01/13), we would prefer to proceeding this PR as is. Further changes can be added in a follow-up PR then. Thanks!

SamuelTrahanNOAA:
> Further changes can be added in a follow-up PR then. Thanks!

That's just as well, since it'll give GDIT time to experiment with optimization options. It may be the key to reducing the 60s timestep's node count.

jkbk2004:
@SamuelTrahanNOAA are you working today? can we combine in #1558 to this pr? FYI @DusanJovic-NOAA

jkbk2004:
We can start working on this pr.

SamuelTrahanNOAA:
I'll merge it in now and do a little test to make sure everything still works.

SamuelTrahanNOAA:
This branch was also not up to date with develop. I've updated it in my local clone, and merged #1558, and I'm testing it now.

SamuelTrahanNOAA:
@DusanJovic-NOAA Please check that all of the changes from #1558 are in here.

DusanJovic-NOAA:
> @DusanJovic-NOAA Please check that all of the changes from #1558 are in here.

Yes, they are. Thanks.

jkbk2004:
Cool! I will run RTs from hera.

SamuelTrahanNOAA:
> Cool! I will run RTs from hera.

Remember to regenerate hafs baselines for Intel. No other tests should change results.

I'm running hera.gnu and hera.intel presently, too. I'll let you know the results when I have them.

SamuelTrahanNOAA:
~~In my gnu tests, control_p8, control_stochy, and control_c48 changed results.~~

EDIT: Sorry, there was an error in how I ran the gnu tests. (I forgot RT_COMPILER=gnu) I'm starting them over now.

SamuelTrahanNOAA:
The FV3_GFS_v16_fv3wam failed to build because the compiler segfaulted.

```
/tmp/ifortBe4yWD.i90: catastrophic error: **Internal compiler error: segmentation violation signal raised** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
err_read_mod_file compilation aborted for /scratch2/BMC/wrfruc/Samuel.Trahan/gsl-pr/supafast/FV3/ccpp/physics/physics/rte-rrtmgp/tests/clear_sky_regression.F90 (code 1)
```

I'll resubmit and see if it succeeds.

EDIT: that was with the Intel compiler.

SamuelTrahanNOAA:
The wam build succeeded the second time, and both of its regression tests passed.

SamuelTrahanNOAA:
Success! I generated the hera.intel hafs baselines, and copied the baselines for all other intel tests. The hera.intel and hera.gnu both reproduced their baselines.

Please start your tests, if you haven't already.

BrianCurtis-NOAA:
***Automated RT Failure Notification***
Machine: hera
Compiler: intel
Job: RT
[RT] Repo location: /scratch1/NCEPDEV/nems/emc.nemspara/autort/pr/1181503486/20230110003015/ufs-weather-model
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn_wav 127 failed in check_result failed 
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn_wav 127 failed in run_test failed 
Please make changes and add the following label back: hera-intel-RT

BrianCurtis-NOAA:
***Automated RT Failure Notification***
Machine: orion
Compiler: intel
Job: RT
[RT] Repo location: /work/noaa/nems/emc.nemspara/autort/pr/1181503486/20230109183009/ufs-weather-model
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn_wav 127 failed in check_result failed 
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn_wav 127 failed in run_test failed 
Please make changes and add the following label back: orion-intel-RT

BrianCurtis-NOAA:
***Automated RT Failure Notification***
Machine: gaea
Compiler: intel
Job: RT
[RT] Repo location: /lustre/f2/pdata/ncep/emc.nemspara/autort/pr/1181503486/20230110003006/ufs-weather-model
[RT] Error: Test rap_control_dyn32_phy32 102 failed in run_test failed 
[RT] Error: Test hafs_regional_atm_ocn 115 failed in check_result failed 
[RT] Error: Test hafs_regional_atm_ocn 115 failed in run_test failed 
[RT] Error: Test hafs_regional_atm_ocn_wav 117 failed in check_result failed 
[RT] Error: Test hafs_regional_atm_ocn_wav 117 failed in run_test failed 
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn 124 failed in check_result failed 
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn 124 failed in run_test failed 
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn_wav 126 failed in check_result failed 
[RT] Error: Test hafs_regional_storm_following_1nest_atm_ocn_wav 126 failed in run_test failed 
[RT] Error: Test hafs_regional_datm_cdeps 129 failed in check_result failed 
[RT] Error: Test hafs_regional_datm_cdeps 129 failed in run_test failed 
Please make changes and add the following label back: gaea-intel-RT

jkbk2004:
@SamuelTrahanNOAA @BrianCurtis-NOAA we need to add new BL_DATE. But I like to confirm those 4 failed cases on gaea.

jkbk2004:
cray seems to respond differently with removing -xSSE4.2 in hycom cmake.

SamuelTrahanNOAA:
Please confirm the problem first. However, if it is the problem, is there a way to use -xSSE4.2 only on GAEA?

SamuelTrahanNOAA:
@jkbk2004 @BrianCurtis-NOAA Please retest GAEA. I think I turned on `-xSSE4.2` on GAEA only.

jkbk2004:
Ok! I am adding new BL_DATE=20230110.

BrianCurtis-NOAA:
Acorn (Intel): hafs_regional_storm_following_1nest_atm_ocn_wav also failed to run in the RT create stage.

SamuelTrahanNOAA:
> Acorn (Intel): hafs_regional_storm_following_1nest_atm_ocn_wav also failed to run in the RT create stage.

What was the error?

If Cactus/Dogwood succeed, so should Acorn, unless there are mismatches in module versions or some software was installed incorrectly.

BrianCurtis-NOAA:
I was able to get the Acorn RT created, and the RT's all pass except for control_atmwav, the atmf000.nc file does not reproduce.. I'm looking into it still. This is just FYI.

jkbk2004:
Still something to discuss about the issue on gaea. it changes results on cpld_control_c192_p8. regional_atmaq_debug crashes with forrtl: error (78): process killed (SIGTERM). I think we can update cpld_control_c192_p8 baseline. @BrianCurtis-NOAA what about regional_atmaq_debug ? turn it off on gaea and tackle down later? I am double checking with develop branch.

jkbk2004:
@BinLiu-NOAA note that we need a bit more time to move on to #1544. Is that ok?

BinLiu-NOAA:
> @BinLiu-NOAA note that we need a bit more time to move on to #1544. Is that ok?

Hi @jkbk2004, that should be fine. As long as we can complete the PR #1544 saying this Friday or by latest next Tuesday, we are still kind of on schedule. Thanks!

BrianCurtis-NOAA:
@jkbk2004 Did the atmaq hit the wall clock limit?

jkbk2004:
@BrianCurtis-NOAA you are right: CANCELLED AT 2023-01-11T14:25:16 DUE TO TIME LIMIT. I think we can turn it off.

BrianCurtis-NOAA:
> @BrianCurtis-NOAA you are right: CANCELLED AT 2023-01-11T14:25:16 DUE TO TIME LIMIT. I think we can turn it off.

Thats fine for Gaea.

jkbk2004:
Ok! the solution is to update cpld_control_c192_p8 baseline and turn off  regional_atmaq_debug on gaea. I need a bit more time to update the log on gaea.

DeniseWorthen:
@jkbk2004 Do I understand---only the non-hafs c192 baseline changes, and only on Gaea? Why ?

jkbk2004:
> @jkbk2004 Do I understand---only the non-hafs c192 baseline changes, and only on Gaea? Why ?

Yes, that was my question. only supposed to change hafs_regional_storm_following_1nest_atm_ocn_wav but cpld_control_c192_p8 changes on gaea.

DeniseWorthen:
@jkbk2004 But it sounds like you are just updating the c192 baseline on gaea instead of trying to figure out why it changed?

jkbk2004:
I am ncdiff-ing. @SamuelTrahanNOAA we need a conclusion here.

DeniseWorthen:
@jkbk2004 Where are run directories located on gaea. I used to find them in /lustre/f2/scratch/emc.nemspara

jkbk2004:
/lustre/f2/scratch/Jong.Kim/FV3_RT/REGRESSION_TEST_INTEL/cpld_control_c192_p8 is new one that is different from current baseline /lustre/f2/pdata/ncep_shared/emc.nemspara/RT/NEMSfv3gfs/develop-20230110/INTEL

SamuelTrahanNOAA:
> Yes, that was my question. only supposed to change hafs_regional_storm_following_1nest_atm_ocn_wav but cpld_control_c192_p8 changes on gaea.

First off, *all* of the hafs tests are supposed to change results. However, nothing I did could have changed tests that don't use hycom.

I would like you to do one of these:

1. Test Dusan's PR #1558 on GAEA
2. Reverse-merge #1558 from this PR's branch and retest on GAEA.

This PR is  needed for the HAFS code freeze, so it does need to be merged as soon as possible, unless @BinLiu-NOAA or @ZhanZhang-NOAA can confirm that the implementation timeline has changed.

DeniseWorthen:
@jkbk2004 Please check and confirm that you are not using 32Bit=ON for the compile. In this job /lustre/f2/pdata/ncep/Jong.Kim/rt-1554/tests, the Regression_test.gaea.log shows you compiled with 32Bit=ON:

```
Compile 001 elapsed time 1291 seconds. -DAPP=S2SWA -D32BIT=ON -DCCPP_SUITES=FV3_GFS_v17_coupled_p8,FV3_GFS_cpld_rasmgshocnsstnoahmp_ugwp -DMPI=ON -DCMAKE_BUILD_TYPE=Release -DMOM6SOLO=ON
```

jkbk2004:
> @jkbk2004 Please check and confirm that you are not using 32Bit=ON for the compile. In this job /lustre/f2/pdata/ncep/Jong.Kim/rt-1554/tests, the Regression_test.gaea.log shows you compiled with 32Bit=ON:
> 
> ```
> Compile 001 elapsed time 1291 seconds. -DAPP=S2SWA -D32BIT=ON -DCCPP_SUITES=FV3_GFS_v17_coupled_p8,FV3_GFS_cpld_rasmgshocnsstnoahmp_ugwp -DMPI=ON -DCMAKE_BUILD_TYPE=Release -DMOM6SOLO=ON
> ```

Oh! let me check

DeniseWorthen:
I checked out Sam's 9d449cbd and ran all cpld tests on Gaea w/ no changes. All baselines passed. I did not run the HAFS tests.

DusanJovic-NOAA:
I also ran #1558 branch code on Gaea and cpld_control_c192_p8 test passed against the current baseline.

jkbk2004:
Yeah, it's running on my side. Hopefully, we can finish today.

jkbk2004:
@SamuelTrahanNOAA I like to confirm one thing. this pr only changes affect hafs_regional_storm_following_1nest_atm_ocn_wav. no imapct on all other hafs tests, right?

jkbk2004:
@SamuelTrahanNOAA @BrianCurtis-NOAA All tests are done. the gaea issue was my mistake. All ok now.  This pr affects only on hafs_regional_storm_following_1nest_atm_ocn_wav.  We are ready to merge the pr.

BrianCurtis-NOAA (changes requested):
> @SamuelTrahanNOAA @BrianCurtis-NOAA All tests are done. the gaea issue was my mistake. All ok now. This pr affects only on hafs_regional_storm_following_1nest_atm_ocn_wav. We are ready to merge the pr.

@jkbk2004 We haven't confirmed from @SamuelTrahanNOAA why the hafs tests don't fail comparing to previous baselines. We need that to continue.

BrianCurtis-NOAA:
Compared on WCOSS2 to previous baselines and hafs_regional_atm_ocn, hafs_regional_atm_ocn_wav, hafs_regional_storm_following_1nest_atm_ocn, and hafs_regional_storm_following_1nest_atm_ocn_wav all do not compare sucessfully. They all required new baselines.

jkbk2004:
> Compared on WCOSS2 to previous baselines and hafs_regional_atm_ocn, hafs_regional_atm_ocn_wav, hafs_regional_storm_following_1nest_atm_ocn, and hafs_regional_storm_following_1nest_atm_ocn_wav all do not compare sucessfully. They all required new baselines.

@BrianCurtis-NOAA cmake/Intel.cmake lines 23-27 same as develop branch. so all cases reproduce except hafs_regional_storm_following_1nest_atm_ocn_wav

BinLiu-NOAA:
@jkbk2004, @BrianCurtis-NOAA, there is a HYCOM build option change taking out "-xSSE4.2". So some other HAFS HYCOM coupled RTs could also get result changing.

jkbk2004:
> @jkbk2004, @BrianCurtis-NOAA, there is a HYCOM build option change taking out "-xSSE4.2". So some other HAFS HYCOM coupled RTs could also get result changing.

Let me confirm again with some hafs tests on hera. Weird thing is all other hafs cases not affacted.

DeniseWorthen:
@BinLiu-NOAA Doesn't the removal of -xSSE4.2 only happen for Gaea? Meaning that testing on Hera won't tell you anything about how the tests will compare on Gaea?

BinLiu-NOAA:
> @BinLiu-NOAA Doesn't the removal of -xSSE4.2 only happen for Gaea? Meaning that testing on Hera won't tell you anything about how the tests will compare on Gaea?

@DeniseWorthen, the removal of "-xSSE4.2" is meant to happen to at least on WCOSS2 (maybe also on Hera/Orion/Jet). I think @SamuelTrahanNOAA made Gaea to keep using "-xSSE4.2" (since it failed on Gaea after taking it out).

DeniseWorthen:
@BinLiu-NOAA Thanks, I guess I had it reversed!

SamuelTrahanNOAA:
What is the status of this PR?

BinLiu-NOAA:
@SamuelTrahanNOAA, I think we are waiting for @jkbk2004 to double check if the HAFS HYCOM coupling RTs also got result changing on Hera (which may or may not be the case). But, otherwise, it looks all the regression tests have gone through.

jkbk2004:
Confirms hafs_regional_atm_ocn is not changing on hera. So all hafs cases on hera/orion/gaea/jet/cheyenne same results except hafs_regional_storm_following_1nest_atm_ocn_wav. So, we can move one at least if "-xSSE4.2" handling behaves as expected on wcoss2. @BinLiu-NOAA what do you think?

BinLiu-NOAA:
@jkbk2004, @BrianCurtis-NOAA, @SamuelTrahanNOAA, we would like to remove using "-xSSE4.2" option for all platforms (except on Gaea, because it fails). Even if it turns out to change results, we can just accept the fact of result changing for those HAFS-HYCOM coupled RTs. Thanks!

jkbk2004:
I agree. we can investigate more across machines. I will create an issue.

jkbk2004:
@BrianCurtis-NOAA let's go ahead to merge.

BinLiu-NOAA:
> I agree. we can investigate more across machines. I will create an issue.

@jkbk2004, Thanks for confirming all HAFS HYCOM coupling RTs (except the one except hafs_regional_storm_following_1nest_atm_ocn_wav which modified to use the FASTER build option) can reproduce the previous baseline results on hera/orion/gaea/jet/cheyenne. And the HAFS HYCOM RTs are expected to change results on WCOSS2 (since on WCOSS2 we are not recommended to this -xSSE4.2 option). And that's why we are taking out this -xSSE4.2 option in this PR. So, for me, all the result changing the corresponding HAFS RTs are expected/understandable. Thus, I do not see any issue to be concerned/created. Thanks!

jkbk2004:
created an issue https://github.com/ufs-community/ufs-weather-model/issues/1565

jkbk2004 (approved):

BrianCurtis-NOAA (approved):

PR #2316, opened 2024-06-10
Ss 160

RatkoVasic-NOAA:
Quick fix to upgrade Derecho to spack stack 1.6.0


BrianCurtis-NOAA:
Why does this not use `load("ufs_common")` ?

RatkoVasic-NOAA:
> Why does this not use `load("ufs_common")` ?

IDK, it was inherited from before, Maybe someone else knows why.

jkbk2004:
merged with #2283

PR #2326, opened 2024-06-13
sync with head of NOAA-EMC UPP develop

SamuelTrahanNOAA:
## Commit Queue Requirements:

- [x] Fill out all sections of this template.
- [x] All sub component pull requests have been reviewed by their code managers (still hoping for more reviews, though)
- [x] Run the full Intel+GNU RT suite (compared to current baselines) on either Hera/Derecho/Hercules
- [x] Commit 'test_changes.list' from previous step
---

## Description:

1. Sync FV3/upp with the head of its authoritative branch.
2. Update library g2tmpl to version 1.13.0
3. Update library g2 to version 3.5.1
4. Inline post FV3/io/post_FV3.F90 is updated for the new UPP.
5. Update the UPP control files for global, rrfs and hafs used in inline post in regression tests.
6. FV3's spack.yaml has explicit UPP dependencies (g2, g2tmpl, etc.) and all libraries have exact version numbers. These version numbers match the top of ufs-weather-model's develop branch for hercules.

Syncing to the top of FV3/upp brings in many bug fixes and new features. Users must update their build systems and static files due to this update.

### After this update, users must...

1. Update to g2tmpl version 1.13.0.
2. Update to g2 version 3.5.1
3. Ensure ufs-weather-model/modulefiles, FV3/io/post_FV3.F90, and FV3/upp are all updated in their forks.
4. Regenerate postxconfig*.txt files if running the inline post.

### The reasons users must update are...

1. Some variables change from GRIB template 4.8 to 4.9. This requires a newer version of g2tmpl. Also, it requires changes to postxconfig*.txt files if you use those variables. GRIB records for those variables will change to, so it will affect downstream users who use those fields.
2. As a cleanup measure, variable names for number concentration on hybrid levels no longer have the "GSL_" prefix. Your postxconfig*.txt files must update to match, if you use those variables.
3. UPP's EBB variable changed from 2D to 4D. This requires a change to FV3/io/post_FV3.F90 to send CCPP's ebu_smoke to UPP's EBB. That change is not backward-compatible to older versions. Both post_FV3.F90 and UPP must update at the same time.

### Commit Message:

```
* UFSWM - update g2tmpl library, g2 library, UPP hash. Sync parm files with new UPP version.
  * FV3 - update inline post for UPP, point UPP submodule to head of develop, revise spack.yaml
```

### Priority:
* Normal

### UFSWM:
- #2322 
- #2321

### Sub component Pull Requests:
[...]

PR #2265, opened 2024-05-03
Feature/detect frontera

benjamin-cash:
## Commit Queue Requirements:

- [x] Fill out all sections of this template.
- [ ] All sub component pull requests have been reviewed by their code managers.
- [ ] Run the full Intel+GNU RT suite (compared to current baselines) on either Hera/Derecho/Hercules
- [ ] Commit 'test_changes.list' from previous step
---
### Commit Message:

```
* UFSWM - This only affect detect_machine.sh
  
```

### Priority:

* Normal

### UFSWM:

* Closes # https://github.com/NOAA-EMC/global-workflow/issues/2570 
(Note that I did not realize ufs-weather-model was the authoritative copy of this script when I created the issue)

### Sub component Pull Requests:

* None

---
### Regression Test Changes (Please commit test_changes.list):

* No Baseline Changes.

### Input data Changes:

* None.

## Testing Log:
- RDHPCS
  - [ ] Hera
  - [ ] Orion
  - [ ] Hercules
  - [ ] Jet
  - [ ] Gaea
  - [ ] Derecho
- WCOSS2
  - [ ] Dogwood/Cactus
  - [ ] Acorn
- [ ] CI
- [ ] opnReqTest (complete task if unnecessary)

PR #2335, opened 2024-06-21
Restore error checking in regression test system. (Combined PR#2357 and PR#2265)

### LINKED ISSUE #2330: modify report of test failures to clearly indicate when a test failed to compare because it did not run
## Description

In PR #2290, many tests were expected to fail comparison because of the code changes. Multiple tests however failed because they did not actually run. Since developers were expecting test failures, they were not alerted to the fact that the tests failed because of seg-faults. This leads to delays during the code commit process, since fixing this type of test failure requires debugging etc. This type of failure needs to be clearly differentiated from cases where the tests failed because the comparisons failed. 

Secondly, looking at the failure in this PR for the ``cpld_control_gfsv17_intel`` case, the log reports

```
* TEST cpld_control_gfsv17_intel: FAILED: UNABLE TO COMPLETE COMPARISON
-- LOG: /scratch1/BMC/gmtb/Grant.Firl/ufs-weather-model-test/tests/logs/log_hera/run_cpld_control_gfsv17_intel.log
```

In fact, in this case, all the comparisons were made (and failed). The comparison was completed so the log message is misleading.

## Solution

A test failure should clearly indicate why it failed.

SamuelTrahanNOAA:
## Commit Queue Requirements:

- [x] N/A No subcomponents. ~~All sub component pull requests have been reviewed by their code managers.~~
- [ ] Run the full Intel+GNU RT suite (compared to current baselines) on either Hera/Derecho/Hercules
- [ ] Commit 'test_changes.list' from previous step
---
## Description:
The regression test system ignores all errors in all jobs. A job where fv3.exe crashed is the same as a job where it ran, and produced different results. Also, compilation job errors are ignored. This leads to several problems:

1. Test jobs are executed even if the compile job fails. (This bug is restricted to the Rocoto-based system.)
2. Jobs with prerequisites (such as restart tests) run even if their prerequisite fails.
3. A job that failed to copy input data or had syntax errors won't be caught until the entire workflow completes.
4. Temporary system issues require rerunning the entire workflow instead of only affected jobs.

In this new version of the regression test system:

1. Errors are caught, and result in the metascheduler considering the job as failed.
2. Dependencies are honored; if a job fails, anything that depends on it won't run.
3. *Jobs that run to completion, but have changed results are considered to have succeeded. This behavior is unchanged.*

### New Self-Test System

A new `tests/error-test.conf` has several jobs to test whether the rt.sh fails and succeeds properly. This is a semi-automated self-test. It contains:

1. `fail_to_copy` - This test will fail in run_test.sh before running the job_card. That means ecflow will be unable to submit the job, and other metaschedulers will see the job start and abort with exit status 1
2. `fail_to_run` - A test that fails inside the job_card. All metaschedulers should see the job abort with exit status 1.
3. `fail_to_compile` - A compile job that will always fail to compile.
4. `dependency_unmet` - depends on the `fail_to_compile`, so it should not be submitted.
5. `atm_dyn32` and `control_c48.v2.sfc` - These should succeed and match the baselines.
6. `atm_dyn64` and `control_c48` - Should succeed, but not match the baselines, because they use 64-bit dynamics instead of 32-bit.

All tests are variations on the control_c48 since it's one of the cheapest and oldest tests in rt.conf, but is unlikely to go away. (The control_c48 tests a core functionality: super-low-res GFS.)

### Commit Message:

```
* UFSWM - restore error checking to regression test system and add a self-test suite
```

### Priority:
* Normal

### UFSWM:
* Closes #2330

N/A

This branch correctly detects failing tests. That means the regression tests will keep failing until this bug fix is merged:
- #2362 

---
* No Baseline Changes.

### Input data Changes:
* None.

### Library Changes/Upgrades:
* No Updates
  
---

## Testing Log:
- RDHPCS
  - [x] Hera
  - [x] Orion
  - [x] Hercules
  - [x] Jet
  - [x] Gaea
  - [x] Derecho
- WCOSS2
  - [x] Dogwood/Cactus
  - [ ] Acorn
- [x] CI


SamuelTrahanNOAA:
Pinging  @DeniseWorthen who authored the relevant issue. Also, @DusanJovic-NOAA who authored the original regression test system.

SamuelTrahanNOAA:
I've been testing this on top of #2326. The regression test has proven itself completely unusable for development due to the lack of error checking. Updating UPP and modulefiles required many changes that caused subsets of the tests to fail. A regression test system that is unable to differentiate between a test with changed results, and a test that could not run at all, is not useful for development.

jkbk2004:
@SamuelTrahanNOAA Can you follow up to clean the super-linter complaint ?

SamuelTrahanNOAA:
After updating this branch, I'm getting out-of-memory errors from some jobs on Hera when using Rocoto.

SamuelTrahanNOAA:
> After updating this branch, I'm getting out-of-memory errors from some jobs on Hera when using Rocoto.

The jobs succeeded on the second attempt. This may have been a temporary system issue.

jkbk2004:
@SamuelTrahanNOAA can you sync up branch? we can start working on this pr. We also want to combine #2265 and #2357 to this pr as well.

BrianCurtis-NOAA:
@jkbk2004 There are a TON of FIXME in these code changes. This should not be combined into anything until those are addressed.

jkbk2004:
@BrianCurtis-NOAA I agree we will do enough test of this new functionality of error checking.  #2265 and #2357 are very minor PRs. Lets see how test goes. @zach1221 @FernandoAndrade-NOAA let's confirm how error checking runs on each machine.

SamuelTrahanNOAA:
I've only tested this with Rocoto. For other workflow managers, I used educated guesses when modifying the scripts. The ecFlow system has logic and scripts outside of the tests/*.sh files. I don't know how that'll break things.

zach1221:
> I've only tested this with Rocoto. For other workflow managers, I used educated guesses when modifying the scripts. The ecFlow system has logic and scripts outside of the tests/*.sh files. I don't know how that'll break things.

@SamuelTrahanNOAA can you please sync up your PR? @FernandoAndrade-NOAA and I can then test ecflow across the rdhpcs.

SamuelTrahanNOAA:
You should test this on one machine before you do all machines. Testing this is a manual process, since you need to confirm the jobs are failing as intended. If the workflow completes, then the error checking is still broken.

SamuelTrahanNOAA:
Some regional debug jobs are failing frequently due to this bug:

- https://github.com/ufs-community/ufs-weather-model/issues/2360
- https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/346

I strongly recommend you reverse merge the changes that caused the bug unless a fix can be found soon. Alternatives aren't great: disable several debug jobs, or let the jobs fail repeatedly.

zach1221:
Some results from testing with ecflow on Hercules, without -c option.
rrfs_v1beta_failing test completed but failed to match baselines. I believe this test is supposed to fail at runtime.

Compile_atm_faster_dyn32_intel and tests control_p8_faster  and regional_control_faster  completed but did not match baselines.

compile_hafw_intel failed at runtime and test never ran.

SamuelTrahanNOAA:
It sounds like we're most of the way there. This is the only unexpected behavior:

>  rrfs_v1beta_failing test completed but failed to match baselines. I believe this test is supposed to fail at runtime.

Can you point me to the log file for that job?

FernandoAndrade-NOAA:
From my sample run on Hera with ecflow without -c:
```
FAILED: UNABLE TO COMPLETE COMPARISON -- TEST 'rrfs_v1beta_failing_intel' [, ]( MB)

PASS -- COMPILE 'atm_faster_dyn32_intel' [18:12, 10:49] ( 1 remarks )
FAILED: UNABLE TO COMPLETE COMPARISON -- TEST 'control_p8_faster_intel' [, ]( MB)
FAILED: UNABLE TO COMPLETE COMPARISON -- TEST 'regional_control_faster_intel' [, ]( MB)

FAILED: UNABLE TO FINISH COMPILE -- COMPILE 'hafsw_intel' [, ]
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_atm_thompson_gfdlsf_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_atm_ocn_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_atm_wav_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_atm_ocn_wav_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_1nest_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_telescopic_2nests_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_global_1nest_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_global_multiple_4nests_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_specified_moving_1nest_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_storm_following_1nest_atm_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_regional_storm_following_1nest_atm_ocn_intel' [, ]( MB)
SKIPPED: ASSOCIATED COMPILE FAILED -- TEST 'hafs_global_storm_following_1nest_atm_intel' [, ]( MB)
```
It may be a good idea to eventually add a differentiation in the failure message for tests like those under the `atm_faster_dyn32_intel` compile that indicate the tests completed fine, and that results are not matching.  Unable to complete comparison sounds like there is an issue with the comparison itself in my opinion, but that's open to discussion.

Paths for confirmation of intended behavior:
`/scratch1/NCEPDEV/stmp2/Fernando.Andrade-maldonado/FV3_RT/rt_2652290`
`/scratch2/NAGAPE/epic/Fernando.Andrade-maldonado/regression-tests/wm/2335/ufs-weather-model/tests`

SamuelTrahanNOAA:
To know what went wrong, I need to see the log file of the stdout and stderr from the rrfs_v1beta_failing_intel job.

zach1221:
> To know what went wrong, I need to see the log file of the stdout and stderr from the rrfs_v1beta_failing_intel job.

@SamuelTrahanNOAA sure, the out/err logs are here: /work2/noaa/stmp/zshrader/stmp/zshrader/FV3_RT/rt_708829/rrfs_v1beta_failing_intel

SamuelTrahanNOAA:
I need to see the full log from the entire job. The "out" and "err" files only contain the stdout and stderr from the "job_card" script. Whatever is going wrong is in run_tests.sh, whose logs are not in that directory.

zach1221:
> I need to see the full log from the entire job. The "out" and "err" files only contain the stdout and stderr from the "job_card" script. Whatever is going wrong is in run_tests.sh, whose logs are not in that directory.

Ok, well there's the RT log here (/work/noaa/nems/zshrader/hercules/rt-2335/ufs-weather-model/tests/logs)
And the run_test logs here (/work/noaa/nems/zshrader/hercules/rt-2335/ufs-weather-model/tests/logs/log_hercules)

SamuelTrahanNOAA:
I see the problem now. The "if" statement I added to abort the job isn't executed with ecflow. To test ecflow, we actually need two failing test jobs:

1. A job that fails while setting variables and copying inputs.
2. A job that fails inside the job_card

That's because, with ecflow, all of this is done on the ecflow node, not the job:

1. Set all variables.
2. Copy inputs.
3. Compare outputs.

I'll update my failures to include two tests, so we can test ecflow properly.

SamuelTrahanNOAA:
I've replaced the "rrfs_v1beta_failing" job with two jobs:

- `rrfs_v1beta_fail_to_copy` = fails in run_test.sh before running/submitting the job_card. For ecFlow, this will fail to submit the job. For other metaschedulers, the failure will happen inside the job. In any case, the workflow shouldn't complete.
- `rrfs_v1beta_fail_to_run` = fails in job_card. All metaschedulers should see this as a job failing, and the workflow shouldn't complete.

Also, I've added `tests/error-test.conf` with a minimal set of tests needed to determine if error checking works.

zach1221:
> I've replaced the "rrfs_v1beta_failing" job with two jobs:
> 
> * `rrfs_v1beta_fail_to_copy` = fails in run_test.sh before running/submitting the job_card. For ecFlow, this will fail to submit the job. For other metaschedulers, the failure will happen inside the job. In any case, the workflow shouldn't complete.
> * `rrfs_v1beta_fail_to_run` = fails in job_card. All metaschedulers should see this as a job failing, and the workflow shouldn't complete.
> 
> Also, I've added `tests/error-test.conf` with a minimal set of tests needed to determine if error checking works.

Ok, it looks like the results are as you expect now, rrfs_v1beta_fail_to_copy fails before a job can be submitted and rrfs_v1beta_fail_to_run fails within the job_card, prior to test completing. The rrfs compile however builds successfully.

SamuelTrahanNOAA:
> The rrfs compile however builds successfully.

The only compilation that should fail is hafsw_intel, because I added the invalid argument `--invalid-argument` to the cmake argument list.

SamuelTrahanNOAA:
Perhaps we should retain this ability to test the error checking?

I could clean up the functionality, and we'll maintain a separate `error-test.conf` file. Then every week or two, you can check if the regression test system is still working.

zach1221:
> Perhaps we should retain this ability to test the error checking?
> 
> I could clean up the functionality, and we'll maintain a separate `error-test.conf` file. Then every week or two, you can check if the regression test system is still working.

I think that makes sense. @BrianCurtis-NOAA @jkbk2004 what do you think?

tests/error-test.conf
```diff
@@ -0,0 +1,26 @@
+
+
+# FIXME: THIS FILE SHOULD NOT BE MERGED TO DEVELOP
```
BrianCurtis-NOAA: do we need these FIXME if we're keeping this now?

tests/fv3_conf/fv3_qsub.IN_acorn
```diff
@@ -30,6 +30,13 @@ export ESMF_RUNTIME_COMPLIANCECHECK=OFF:depth=4
 export ESMF_RUNTIME_PROFILE=ON
 export ESMF_RUNTIME_PROFILE_OUTPUT="SUMMARY"
 
+# FIXME: THIS NEW "IF" BLOCK SHOULD NOT BE MERGED TO DEVELOP
```
BrianCurtis-NOAA: this FIXME?

tests/rt.conf
```diff
@@ -1,3 +1,8 @@
+
+
+# FIXME: CHANGES TO THIS FILE SHOULD NOT BE MERGED TO DEVELOP
```
BrianCurtis-NOAA: Should this file be returned to the develop version?

BrianCurtis-NOAA (changes requested):

tests/rt.conf
SamuelTrahanNOAA: Yes.

BrianCurtis-NOAA:
Maybe EPIC could try running the purposely failing tests as a part of CI on push for PR's ?

tests/fv3_conf/fv3_qsub.IN_acorn
SamuelTrahanNOAA: All the "FIXME" lines will be replaced with explanatory comments.

tests/error-test.conf
SamuelTrahanNOAA: I'll replace it with an explanatory comment.

SamuelTrahanNOAA:
> Maybe EPIC could try running the purposely failing tests as a part of CI on push for PR's ?

The failure tests really only need to run when someone is changing the regression test system itself, or adding new platforms.

SamuelTrahanNOAA:
I've cleaned up the files and renamed the failing tests to more sensible things. They're all variants of control_c48 since it's one of the cheapest and oldest tests.

DusanJovic-NOAA:
@SamuelTrahanNOAA The PR description says that:

> Test jobs are executed even if the compile job fails.

Can you please provide more context for this claim, what kind of compile job failure are ignored such that the tests that depend on the failed compile job are executed. Can you give us specific sequence of commands that triggers this behavior.

I just created simple .conf file, like this:

```
COMPILE | atm_dyn32 | intel | -DAPP=ATM -DCCPP_SUITES=FV3_GFS_v16,FV3_GFS_v16_flake,FV3_GFS_v17_p8,FV3_GFS_v17_p8_rrtmgp,FV3_GFS_v15_thompson_mynn_lam3km,FV3_WoFS_v0,FV3_GFS_v17_p8_mynn,FV3_GFS_v17_p8_ugwpv1 -D32BIT=ON | | fv3 |
RUN | control_p8                                        | - noaacloud                          | baseline |
RUN | control_restart_p8                                | - noaacloud                          |          | control_p8
```

and I introduced a syntax error in one of the source code files to cause compile error, and I see that the two RUN jobs are force aborted, as they should:

```
rt_utils.sh: Starting ECFLOW tasks...
rt_utils.sh: Total number of tasks processed -- 3

ECFLOW Tasks Remaining: 3/3 .........................................................................................
ECFLOW Tasks Remaining: 2/3 Will force aborted state for task /regtest_3622071/control_p8_intel
Will force aborted state for task /regtest_3622071/control_restart_p8_intel

ECFLOW Tasks Remaining: 0/3 rt_utils.sh: ECFLOW tasks completed, cleaning up suite
rt.sh: Generating Regression Testing Log...

REGRESSION TEST RESULT: FAILURE
******Regression Testing Script Completed******
```

SamuelTrahanNOAA:
> Can you please provide more context for this claim, what kind of compile job failure are ignored such that the tests that depend on the failed compile job are executed. Can you give us specific sequence of commands that triggers this behavior.

That bug was specific to the Rocoto-based system, because of the way stdout and stderr redirection were handled in the `<command>` in the xml.

DusanJovic-NOAA:
Ok, then the description should be fixed. Claims like:

> The regression test system ignores all errors in all jobs

are misleading and false. Because this implies that entire regression test system is broken and useless, which is not the case.

We recommend that developers use ecflow instead of rocoto. Ecflow is primary workflow manager used by rt.sh simply , because Rocoto is not supported on all platforms, and is not as well tested as ecflow. I suggested several times that we should in fact get rid of Rocoto because of that.

SamuelTrahanNOAA:
I've said several times that I'm happy to support the Rocoto-based system on some machines, for users who are more comfortable with Rocoto.

zach1221:
@SamuelTrahanNOAA I think we may test this PR next, are you able to sync up your branch and combine PR [2357](https://github.com/ufs-community/ufs-weather-model/pull/2357) sphinx doc and PR [2265](https://github.com/ufs-community/ufs-weather-model/pull/2265) frontera, into this?

SamuelTrahanNOAA:
> @ SamuelTrahanNOAA I think we may test this PR next, are you able to sync up your branch and combine PR [2357](https://github.com/ufs-community/ufs-weather-model/pull/2357) sphinx doc and PR [2265](https://github.com/ufs-community/ufs-weather-model/pull/2265) frontera, into this?

Done. I tested the error-check.txt with Rocoto, but not the rt.conf. I'll leave that to you.

SamuelTrahanNOAA:
@benjamin-cash - Please check this PR to confirm your changes were successfully merged.

uturuncoglu:
@SamuelTrahanNOAA JFYI, I think that frontera module file is corrupted and it has html tags etc. Please use the following one to correct it https://github.com/oceanmodeling/ufs-weather-model/blob/feature/coastal_app/modulefiles/ufs_frontera.intel.lua

SamuelTrahanNOAA:
> @ SamuelTrahanNOAA JFYI, I think that frontera module file is corrupted and it has html tags etc. Please use the following one to correct it https://github.com/oceanmodeling/ufs-weather-model/blob/feature/coastal_app/modulefiles/ufs_frontera.intel.lua

I've copied that in. How does it look now?

uturuncoglu:
@SamuelTrahanNOAA It looks good to me. Thanks for it.

zach1221:
@FernandoAndrade-NOAA @BrianCurtis-NOAA I think we're ready to begin the rest of the tests now.

SamuelTrahanNOAA:
Make sure you run the opnReqTest since this PR changes the regression test system itself. Anything related to testing should be tested.

BrianCurtis-NOAA:
still no Acorn, can skip.

zach1221:
> Make sure you run the opnReqTest since this PR changes the regression test system itself. Anything related to testing should be tested.

@SamuelTrahanNOAA ORTs were run successfully. Logs are posted above for control_p8, regional_control and a cpld_control case.
@jkbk2004 the derecho queue really hasn't moved for me all day. Do we want to skip?

SamuelTrahanNOAA:
> the derecho queue really hasn't moved for me all day. Do we want to skip?

Derecho was down for unplanned maintenance for six hours. You could log in, but your jobs would not run. It came out of maintenance about two hours ago.

zach1221:
Looks like the derecho jobs are going through now.

zach1221:
Ok we should be ready now to proceed with merging.

BrianCurtis-NOAA (approved):

jkbk2004 (approved):

DeniseWorthen:
@SamuelTrahanNOAA Thanks for all your work on this. I just ran a PR where I expected two failures. One was because an input file wasn't in the correct subdirectory in the ``input-data``, and one because the answers were going to be different.

The first of these reported a failure ``FAILED: RUN DID NOT COMPLETE``, which is great. But the second, where the run completed and all the comparisons were made (they were all different), still reported the failure as ``FAILED: UNABLE TO COMPLETE COMPARISON``.  Would that be what you expect?

SamuelTrahanNOAA:
I didn't improve the wording of the comparison message. It would be great if they were more detailed than "unable to complete comparison."

DeniseWorthen:
OK, thanks. I do think the message needs to indicate that it failed comparison, not that the comparison didn't complete . Maybe @BrianCurtis-NOAA can take a look at that.

PR #2654, opened 2025-03-19
Add Spack CI

### LINKED ISSUE #2565: add spack-based package file in new spack subdirectory, to allow us to begin maintaining it
Until now, git was used directly when installing the ufs-weather-model, after dependencies are installed (with hpc-stack or spack-stack).

However, the future is that spack will also be used to install ufs-weather-model, and NCO will not use a git clone command, but will instead use spack to install everything, model and dependencies.

The way we have handled this with other projects has been to add a spack subdirectory at the top level, and to put the package file in there. This version of the package file can then be used for testing, especially testing new capabilities before release. The spack-stack project and main spack repos are updated periodically from this package file.

In order to support this we need to add the package file and some CI files. This will be done by @AlexanderRichert-NOAA .

AlexanderRichert-NOAA:
## Commit Queue Requirements:

---
## Description:
This PR adds a CI workflow, Spack.yml, that builds the UFS Weather Model using a newly added in-repo copy of its Spack recipe (ci/package.py) with the intent that it be modified to include dependency changes and new CMake build options. Spack.yml uses the same custom action, NOAA-EMC/ci-test-spack-package, used in the NCEPLIBS Spack CI workflows. Currently Spack.yml uses OpenMPI but other MPI libraries could be straightforwardly used instead or in addition. Software dependencies are installed through Spack, except for those listed under `spack-externals:`, which is to a) reduce workflow run time and b) reduce the size of the dependency cache (see below).

A few things to note:
 - The Spack.yml workflow automatically creates a GitHub Actions cache with the compiled software dependencies. The whole bundle is about 500MB (~120MB without MAPL; GitHub total limit for a repo's caches is 10GB). With the build cache in place, it takes about 8-9 minutes for the workflow for building ATM to run, and about 24 minutes for S2SWA. Without the build cache, the total run time is ~1 hr or more.
 - Currently the ci-test-spack-package action used in this workflow has not been tested with Intel or any other non-GCC compilers. I expect that some modifications to ci-test-spack-package would be needed in order to support this, but it can certainly be done if/when needed.
 - In order to keep the Spack recipe in step with the code, I have added the ci-check-spack-recipe action to the workflow, which checks that all options listed in the root CMakeLists.txt are reflected in the recipe, ci/package.py. This means that when a new CMake option is added, ci/package.py should be modified to reflect it. This should usually be a low effort change, but I am happy to volunteer as needed to assist developers in updating the recipe.
 - I see no reason not to squash-and-merge this PR.

### Commit Message:
```
* UFSWM - "Add Spack-based CI build workflow"
```

### Priority:
* Normal

### UFSWM:
* Closes #2565

* None

* None

---
## Changes
* No Baseline Changes.

### Input data Changes:
* None.

### Library Changes/Upgrades:
* No Updates
  
---

## Testing Log:
- RDHPCS
  - [ ] Hera
  - [ ] Orion
  - [ ] Hercules
  - [ ] GaeaC5
  - [ ] GaeaC6
  - [ ] Derecho
- WCOSS2
  - [ ] Dogwood/Cactus
  - [ ] Acorn
- [ ] CI


AlexanderRichert-NOAA:
FYI @edwardhartnett

ci/package.py
```diff
@@ -0,0 +1,235 @@
...
+    depends_on("netcdf-c")
+    depends_on("netcdf-fortran")
+    depends_on("sp")
+    depends_on("w3emc")
+    depends_on("esmf@:8.0.0", when="@:2.0.0")
+    depends_on("nemsio", when="@:2.0.0")
+    depends_on("w3nco", when="@:2.0.0")
+    depends_on("bacio@2.4.0:", when="@develop")
+    depends_on("crtm", when="@develop")
+    depends_on("esmf@8.3.0:", when="@develop")
```
BrianCurtis-NOAA: These versions mismatch what's currently used. As more of a questions on understanding, what is this section doing and why do the versions seem to be a mismatch.

BrianCurtis-NOAA (dismissed):

ci/package.py
AlexanderRichert-NOAA: This section is how Spack knows what dependencies/requirements there are for the code. I intend to update these versions, as well as add a README in that directory to clarify maintenance (which I expect to be fairly low effort).

AlexanderRichert-NOAA: I updated the ESMF minimum version, added a minimum version for NetCDF, and added a README that hopefully clarifies things as far as maintaining the recipe.

BrianCurtis-NOAA (dismissed):
I'm OK with this moving forward.

gspetro-NOAA:
@AlexanderRichert-NOAA Could we have a status update on this PR? Is it still planned/needed?

AlexanderRichert-NOAA:
I think this is good to go-- I can un-draftify it, and if appropriate I can go ahead and update the branch.

gspetro-NOAA:
@AlexanderRichert-NOAA Thanks! Could you also push a log with results of running the full RT suite (preferably on Ursa)? We are going to be requiring this for future PRs to reduce the chances of unexpected issues in our final testing.

AlexanderRichert-NOAA:
@gspetro-NOAA would you be willing to make an exception on account of there being no code changes? This PR is purely CI.

BrianCurtis-NOAA:
@gspetro-NOAA There's no need to run the full suite. This is a good candidate to go into any other PR that does change baselines/code.

BrianCurtis-NOAA:
@AlexanderRichert-NOAA should we modify how often this runs in a PR? If it runs after each push of RT log's for example, I feel that's a waste of resources.

AlexanderRichert-NOAA:
There may be a couple ways to limit how often the workflows run. The only issue I've run into before with excluding certain files from triggering re-runs is that if it's a required workflow (which I recommend but it isn't strictly necessary), then a commit than only updates the excluded files won't trigger a re-run, and the code can't be merged. This might be avoidable by also allowing the workflow to be manually triggered but I haven't tested this.

AlexanderRichert-NOAA:
It's also possible to make it so that only one copy of the workflow can run at once, where when a new commit triggers re-runs, the existing jobs get whacked. So at least it avoids having a million copies of the same workflow running at once for various commits.

gspetro-NOAA:
@AlexanderRichert-NOAA I think that last point would be particularly useful so that if we consecutively push logs for different systems, it stops running an old action when something new has been pushed. Is there any way to take advantage of caching? Or does that not work in this particular context?

AlexanderRichert-NOAA:
We can and it does, yes-- under the hood, the custom action that runs the Spack-based installation is using prebuilt Spack packages.

When I get a chance I can look at the run concurrency configuration so it stops running the old jobs.

AlexanderRichert-NOAA:
I added the concurrency configuration, and a follow-on test commit successfully caused the old jobs to be cancelled: https://github.com/ufs-community/ufs-weather-model/actions/runs/17928219080

gspetro-NOAA:
Merged in w/#2913.

PR #2690, opened 2025-04-09
Switch to vector remapping and bilinear interpolation for global coupled configurations

### LINKED ISSUE #2466: Grid imprinting when running C192 atmos with 1/4 deg ocean and seaice
## Description
We can see the imprinting of a grid on the sea ice and ocean. The figure attached depicts the seaice concentration standard deviation of a 30 member  ensemble after a 6 hour forecast.

The initial conditions were created by:
1 - interpolating the gfs ensemble members from operations from C384 to C192 
2 - adding the same ocean and seaice initial conditions for the 30 members
3 - omitting the mediator restarts
oops-seice-stddev

### To Reproduce:
I don't think this is machine dependent, but my test was done on Hercules.
- Use a configuration of the coupled UFS that has FV3, MOM6 and CICE6
- use C192 for FV3 and 1/4 deg for MOM6 and CICE
- bootstrap/omit the mediator restart

### LINKED ISSUE #2667: Implement vector regridding and bilinear mapping to mitigate grid imprint issue
## Description

A combination of vector regridding and bilinear mapping of ATM states->ICE appear to resolve the grid imprint, described in #2466.

The switch to bilinear mapping of ATM states to ICE is a long standing issue, which had not been implemented because of the appearance of "stripes" in the mapping when using bilinear mapping. See Issue https://github.com/NOAA-EMC/CMEPS/issues/38. A bugfix now existing in ESMF8.8+ to resolve the stripes.

## Solution

The bilinear mapping and vector regridding capability exists currently in CMEPS, but neither is currently enabled for UFS. During testing, it was found that in order for them to work, a change was also required to the srcMask value for the ATM. Details are provided below. 

In addition, some changes in the generation of RH within CMEPS were found to be necessary. These changes may impact CESM coupling, so they will first be pushed to ESCOMP for testing.

DeniseWorthen:
## Commit Queue Requirements:

---
## Description:

This PR makes multiple changes in CMEPS related to mapping choices between ATM and OCN or ICE. ATM states mapped to ICE (which ICE uses to calculate its own fluxes) are changed from conservative fraction to bilinear. 

ATM surface surface winds mapped to ICE are switched to vector remapping using [patch recovery](https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node5.html#SECTION05012200000000000000) and ATM-OCN stresses (which are calculated by ATM and sent to OCN) are vector mapped conservatively (with the [caveat](https://earthsystemmodeling.org/docs/release/ESMF_8_6_0/ESMF_refdoc/node5.html#SECTION050121700000000000000) that "using vector regridding with a conservative regrid method will not necessarily produce vectors whose components are conservative").

The use of vector remapping is controlled by a configuration attribute added to the appropriate ufs.configure using ``mapuv_with_cart3d``. By default, vector remapping will be applied for all global coupled cases (``mapuv_with_cart3d=.true.``). 

In order to enable these mapping changes, the srcMaskValue for the ATM must also be changed from ``1`` to ``spval``. This means that the ATM is unmasked when mapping these states to other components.

See Issue #2667 for more details on testing these changes.

Also note: If this PR is merged prior to the update to ESMF8.8 (which I believe is safe), B4B differences in mediator restart files are expected at the 8.8 update due to the fix of the 'stripes' issue. 

The three primary changes to mapping in this PR are a) mapping of ATM states->ICE bilinearly instead of conservatively b) mapping ATM winds -> ICE with vector remapping and c) mapping the stress from ATM ->OCN with vector remapping. The following shows the current develop branch (left), this feature branch (middle) and the difference for representative fields for these three changes after 6 hours for the ``cpld_control_p8`` test.

ATM T-bot:
Screenshot 2025-04-30 at 5 55 46 PM

ATM u-winds:
Screenshot 2025-04-30 at 5 56 28 PM

Merged surface stress:
Screenshot 2025-04-30 at 6 04 16 PM

### Commit Message:

```
* UFSWM - add configuration attribute to toggle vector remapping in CMEPS
  * CMEPS - set srcMaskValue for ATM to be spval; use bilinear mapping of ATM states -> ICE; enable vector remapping of ATM wind and OCN surface stress 
```

### Priority:

* High: Needed for GFS and Marine DA

### UFSWM:

* Closes #2667
* Closes #2466

* CMEPS: https://github.com/NOAA-EMC/CMEPS/pull/136

* None

---

* PR Updates/Changes Baselines. 
* All ``cpld`` tests need new baselines. 
* An additional 3 tests also need new baselines: 

1.  atm->wav mapping has been switched to bilinear (``atmwav_control_noaero_p8 intel``)
2. atm->ice mapping has been switched to bilinear/uv3d (``atm_ds2s_docn_pcice``)
3. dependent test has changed fields (``atm_ds2s_docn_dice``)

### Input data Changes:

* None.

### Library Changes/Upgrades:

* No Updates
  
---

## Testing Log:
- RDHPCS
  - [x] Hera
  - [x] Orion
  - [x] Hercules
  - [x] GaeaC6
  - [x] Derecho
- WCOSS2
  - [x] Dogwood/Cactus
  - [x] Acorn
- [x] CI


DeniseWorthen:
On C6, I created a baseline and tried to run against it and I got some failures. Specifically, some of the mediator restart are not B4B (but all other files are). I also got a failure in comparison for all files in ``cpld_decomp_p8_intel`` test. I need to investigate, since these are not expected failures.

These are the cases which fail against the generated baseline
```
cpld_mpi_gfsv17 intel
cpld_2threads_p8 intel
cpld_decomp_p8 intel
cpld_mpi_p8 intel
cpld_mpi_pdlib_p8 intel
```

Of these, all but the decomp test fail only in the mediator restart comparison, generally at 1 or 2 points. The values which differ lead me to believe this is the "stripes" bug, but I will confirm.

DeniseWorthen:
I repeated my test of creating a baseline w/ this branch and running against it on Hera. The same files as above fail to compare, in the same ways. The mpi & 2thread fail only in the mediator restart files. The decomp failure is everywhere. 

To verify the mediator restart failures were due to the stripes bug, I repeated my test using a special branch I created w/o aerosols and using 8.8 on C6 (I don't want to faff around w/  the GOCART/MAPL/ESMF playground fight right now). In that case, only the decomp test failed. The decomp test is a "real" failure, the other failures are non-impactful and due only to the stripes bug.

If the decomp failure can be solved, I believe the path forward for this PR under 8.6 is to temporarily remove the mediator restart files from the comparison lists of the affected tests.

NickSzapiro-NOAA:
"patch" has a larger stencil, right? I don't see how CMEPS does domain decomposition haloes

DeniseWorthen:
It shouldn't need to. CMEPS never cares about anyone's halos.

DeniseWorthen:
For the decomp test (on C6, 8.8), the differences first appear in two export fields from the ATM: ``atmImp_Sa_z`` and ``atmImp_Sa_tbot`` on the left edge of tile 4 (i=1, j=25,26,27) at timestep=58320 (51 x 720 timesteps). Only one of these points (i=1,j=25) is a fractional land point (0.9995).

DeniseWorthen:
I've done various debugging on this issue so far (debug, debug+gnu, hera, c6 etc). The decomp test always fails in the same way.

In fact, I've turned on field dumping in ATM (``DumpFields = true``) and verified that before the fields are ever transferred to CMEPS, differences first appear on Tile4 at 3 points in the two export fields height_lowest and temp_lowest.

DeniseWorthen:
I may have a lead on why the test is failing. Tile 4 actually sits on its side, so that the points that are not B4B are actually near the Kamchatka Peninsula. I think there may be some ifrac/ofrac changes that are triggering the issue.

DeniseWorthen:
@DusanJovic-NOAA @junwang-noaa Would one of you have time to see if you can figure out what is happening to break the decomp test (cpld_decomp_p8_intel)? The branch I've been using to test (which uses 8.8 and no aerosols) is https://github.com/DeniseWorthen/ufs-weather-model/tree/feature/imprint_noaero

I've confirmed that the same branch w/o the CMEPS changes does pass the decomp test, but haven't been able to debug the failure of the two export fields to reproduce when they are exported from ATM (even before CMEPS sees them).

DusanJovic-NOAA:
> @DusanJovic-NOAA @junwang-noaa Would one of you have time to see if you can figure out what is happening to break the decomp test (cpld_decomp_p8_intel)? The branch I've been using to test (which uses 8.8 and no aerosols) is https://github.com/DeniseWorthen/ufs-weather-model/tree/feature/imprint_noaero
> 
> I've confirmed that the same branch w/o the CMEPS changes does pass the decomp test, but haven't been able to debug the failure of the two export fields to reproduce when they are exported from ATM (even before CMEPS sees them).

I'm not familiar with CMEPS, but I can take a look.

DeniseWorthen:
@DusanJovic-NOAA Thanks for looking....in this case, from my diagnostics (eg turning on the field dumping in ATM), the differences are arising in the export fields from the ATM, t_bot and z_bot. So ATM gets the same fields on import, but exports different values at just 3 points on the edge of tile 4. I don't know why the CMEPS changes are causing that. The export fields come out of the dycore from my understanding.

DusanJovic-NOAA:
When I set mapuv_with_cart3d to false in ufs.configure I get reproducible results. I see that mapuv_with_cart3d is used at two places in esmFldsExchange_ufs_mod to control mapping from atm to ocn using mapconsf_uv3d and atm to ice using mappatch_uv3d. When I comment the atm to ice I get reproducible results with just atm to ocn, but non-reproducible result with atm to ice enabled with mappatch_uv3d. It seems to me that the issue is when mappatch_uv3d is used. So just to test this, I replaced mappatch_uv3d with mapconsf_uv3d for atm - ice remapping, like this:

```
              ! call addmap_from(compatm, fldname, compice, mappatch_uv3d, 'one', 'unset')
              call addmap_from(compatm, fldname, compice, mapconsf_uv3d, 'one', 'unset')
```
and results are reproducible. I know this is probably not the correct setting, but maybe it can help you figure out what is causing the reproducibility issue.

DeniseWorthen:
I believe I have a fix for the decomp failure. I have a missing if statement in my CMEPS branch. I will re-test by creating baselines and running against them, but in a sandbox, fixing my bug allowed the decomp test to reproduce control. This was using 8.6. I still expect the mediator history files to fail comparison until we go to 8.8, as explained above.

DeniseWorthen:
I've tested the bugfix in CMEPS (fixing the missing if-clause) using a 1.9.1 SS feature branch on Hercules by creating a baseline and then running against it. This is using a special branch where I've removed the compilation and use of the CHM component.

The CMEPS fix does resolve the decomp test failure, but now the ``cpld_mpi_gfsv17_intel`` failed (it did not fail previously when testing w/ 8.8).  

However, both the ``cpld_debug_gfsv17_intel`` and ``cpld_debug_pdlib_p8_intel`` failed during the baseline creation, with the seg fault previously noted in SCOTCH when running with the 1.9.1 SS. That may be the cause of the failure in the non-debug test, but I will need to test further.

DeniseWorthen:
I've updated this branch w/ fixes from CMEPS. I ran against a staged baseline on Hera (ESMF8.6), and the only failing tests were due to the mediator restart files not matching (due to stripes bug). All other files were B4B. The failed cases were

```
cpld_2threads_p8_intel
cpld_mpi_gfsv17_intel
cpld_mpi_p8_intel
cpld_mpi_pdlib_p8_intel
```

DeniseWorthen:
@jkbk2004 This PR is now fully tested and ready.

jkbk2004:
@DeniseWorthen We can start working on this pr. You can go ahead to sync up branch.

BrianCurtis-NOAA:
@DeniseWorthen i'm getting the following issue:
```
+ set +eux
atparse.bash: line 82: MAPUV3D: unbound variable
+ '[' 1 -eq 0 ']'
```
I assume that variable is not being set somewhere that's looking for it? maybe it needs to come out of the export_datm_cdeps section of default vars?

THis is coming from cpld_control_c48, but i know this issue will block all the tests from setting up their run dirs.

DeniseWorthen:
That is odd. I ran baselines a total of 3 times w/ this branch and had no failures like that. I'll look in the morning.

DeniseWorthen:
I've verified that all coupled tests now run; thanks for the catch @BrianCurtis-NOAA !

BrianCurtis-NOAA:
WCOSS2 and Acorn done: @FernandoAndrade-NOAA The bug I see on occasion in the -b option might have missed atmwav_control_noaero_p8_intel, double check that you're creating baselines for it if you're using -b.

jkbk2004:
We can start merging process for this pr.

BrianCurtis-NOAA (approved):

jkbk2004 (approved):

PR #2715, opened 2025-04-23
improve coupled model startup cost for waves; Add switch to disable InfoBroadcast in fv3 cap run phase2 (was #2721)

DeniseWorthen:
## Commit Queue Requirements:

- [x] Fill out all sections of this template.
- [x] All sub component pull requests have been reviewed by their code managers.
- [x] Run the full Intel+GNU RT suite (compared to current baselines) on either Hera/Derecho/Hercules
- [x] Commit 'test_changes.list' from previous step
---
## Description:

WW3: Refactor WW3 mesh cap to move call to w3init into Advertise phase

ATM: Add model_configure switch sync_fcst_info_to_wgc, false by default, that will skip calling InfoBroadcast in fv3 cap run phase 2. See https://github.com/NOAA-EMC/fv3atm/issues/955 for details.

Since currently no regression test needs to change any attributes during the run phase, we decided to not add (update) model_configure templates by adding new placeholder in order to avoid breaking various workflows.

### Commit Message:

```
* UFSWM - Remove multigrid logical from scripts
* UFSWM - Add switch to disable InfoBroadcast in fv3 cap run phase2
  * FV3 - Add switch to disable InfoBroadcast in fv3 cap run phase2 
  * WW3 - Refactor mesh cap to move call to w3init into Advertise phase
```

### Priority:

* High: Needed by GFSv17

### UFSWM:

* Partially resolves #2709
* Closes https://github.com/NOAA-EMC/WW3/issues/1410
* Closes https://github.com/NOAA-EMC/WW3/issues/1411
* Closes https://github.com/NOAA-EMC/fv3atm/issues/955
### Sub component Pull Requests:

* FV3: NOAA-EMC/fv3atm#958
* WW3: https://github.com/NOAA-EMC/WW3/pull/1415

### UFSWM Blocking Dependencies:

* None

---
### Regression Test Changes (Please commit test_changes.list):

* No Baseline Changes.

### Input data Changes:

* None.

### Library Changes/Upgrades:

* No Updates
  
---

## Testing Log:
- RDHPCS
  - [x] Hera
  - [x] Orion
  - [x] Hercules
  - [x] GaeaC5
  - [x] GaeaC6
  - [x] Derecho
- WCOSS2
  - [x] Dogwood/Cactus
  - [x] Acorn
- [x] CI
- [x] opnReqTest (complete task if unnecessary)

PR #2721, opened 2025-05-01
Add switch to disable InfoBroadcast in fv3 cap run phase2

### LINKED ISSUE #955: Add switch to disable InfoBroadcast in fv3 cap run phase2
In the [PR](https://github.com/NOAA-EMC/fv3atm/pull/835) that added support for varying length of accumulation buckets, we added code that synchronizes the field bundle attributes from fcst to write grid components in fv3 cap. [Here](https://github.com/NOAA-EMC/fv3atm/blob/develop/fv3/fv3_cap.F90#L1241-L1254). Unfortunately, this introduced undesirable slow down due to the blocking of entire fv3 cap (call to ESMF_InfoBroadcast works across all fv3 cap PETs). At this moment there is no version of InfoBroadcast that will broadcast data across a subset of fv3 atm PETs, specifically across fcst PETs and only one of the write grid component PETs, (the active one).

Since the slow-down is significant in GFS v17 configurations and since GFS v17 does not use bucket lengths that are changing during the run, it was suggested that we introduce a model configure switch that will skip this block of code. The switch will be false by default, so by default that block of code will not be executed.

DusanJovic-NOAA:
## Commit Queue Requirements:

- [ ] All sub component pull requests have been reviewed by their code managers.
- [ ] Commit 'test_changes.list' from previous step
---
## Description:

Add model_configure switch sync_fcst_info_to_wgc, false by default, that will skip calling InfoBroadcast in fv3 cap run phase 2. See https://github.com/NOAA-EMC/fv3atm/issues/955 for details.

Since currently no regression test needs to change any attributes during the run phase, we decided to not add (update) model_configure templates by adding new placeholder in order to avoid breaking various workflows. 

### Commit Message:

```
* UFSWM - Add switch to disable InfoBroadcast in fv3 cap run phase2
  * FV3 - Add switch to disable InfoBroadcast in fv3 cap run phase2
```

### Priority:

* Normal

### UFSWM:

* Closes https://github.com/NOAA-EMC/fv3atm/issues/955

* FV3: NOAA-EMC/fv3atm#958

* None

---

* No Baseline Changes.

### Input data Changes:

* None.

### Library Changes/Upgrades:

* No Updates
  
---

## Testing Log:
- RDHPCS
  - [ ] Hera
  - [ ] Orion
  - [ ] Hercules
  - [ ] GaeaC5
  - [ ] GaeaC6
  - [ ] Derecho
- WCOSS2
  - [ ] Dogwood/Cactus
  - [ ] Acorn
- [ ] CI


JessicaMeixner-NOAA:
@DusanJovic-NOAA  - The Global-workflow uses the model_configure input files from the ufs-weather-model.  If it's not added here I think that means we can't turn it on in GFSv17.  I think we'd also want to update the gfsv17 tests to use this new feature if possible.

JessicaMeixner-NOAA:
> @DusanJovic-NOAA - The Global-workflow uses the model_configure input files from the ufs-weather-model. If it's not added here I think that means we can't turn it on in GFSv17. I think we'd also want to update the gfsv17 tests to use this new feature if possible.

Or did I misunderstand that the default behavior is what we want, so this is on already?

junwang-noaa:
We only need default setting. Once G-W is updated with the UFS WM develop branch with this PR committed, no change is needed.

DusanJovic-NOAA:
Correct. Since the default value of the switch is false, from now on fv3 cap will skip syncing Info from fcst to wrt comp, which is what we need for GFS v17. This is effectively the same as commenting out that block of code that calls ESMF_InfoBroadcst, which was slowing down calling of write grid component run phase until all other currently running write grid groups return to fv3 cap.

jkbk2004:
merged with #2715

Issue #3287, opened 2026-07-08
remove "export" of variables from default_vars.sh and tests


DeniseWorthen:
## Description

The "export" of variables is not required because ``run_test.sh`` is a subshell of ``rt.sh``.  Setting all variables without exporting them will set them in that subshell and not have any impact on subsequent executions of run_test.sh. 

The "export" is long-standing use in the RT framework but interferes w/ the proposed ``dryrun`` option; since the exports are not actually required, they should be removed.

## Solution

Remove "export VARIABLE_NAME" from default_vars.sh and tests.

This solution was proposed by @DusanJovic-NOAA to resolve an issue with the ``dryrun`` option. See https://github.com/ufs-community/ufs-weather-model/pull/3284#issuecomment-4906480258

NickSzapiro-NOAA:
Did we miss some exports @DeniseWorthen like
https://github.com/ufs-community/ufs-weather-model/blob/a45c31fd7fe2905f605cf59c8da171a985273a98/tests/default_vars.sh#L1427

https://github.com/ufs-community/ufs-weather-model/blob/a45c31fd7fe2905f605cf59c8da171a985273a98/tests/default_vars.sh#L1499

https://github.com/ufs-community/ufs-weather-model/blob/a45c31fd7fe2905f605cf59c8da171a985273a98/tests/default_vars.sh#L1516

DeniseWorthen:
Yes, it looks like the GHA only flags ``export`` when it the first word on a line. I didn't catch that, thanks.

I'll make a fix; should be combinable w/ anything.

gspetro-NOAA:
On a related note, are lines like this required anymore? Our pipeline is failing on these lines, which used to have an `export` in front of them, with `line 165: np2: command not found`. https://github.com/ufs-community/ufs-weather-model/blob/a45c31fd7fe2905f605cf59c8da171a985273a98/tests/tests/cpld_control_c1152_v17#L165-L167

BrianCurtis-NOAA:
> On a related note, are lines like this required anymore? Our pipeline is failing on these lines, which used to have an `export` in front of them, with `line 165: np2: command not found`.
> 
> [ufs-weather-model/tests/tests/cpld_control_c1152_v17](https://github.com/ufs-community/ufs-weather-model/blob/a45c31fd7fe2905f605cf59c8da171a985273a98/tests/tests/cpld_control_c1152_v17#L165-L167)
> 
[...]

PR #3289, opened 2026-07-09
remove potential octal traps from default_vars and tests/*; add GHA to guard against re-introduction (#3289) // Derecho bugfix (#3291) // Update CCPP links (#3298)

### LINKED ISSUE #3120: Derecho hanging when using both Rocoto and ecFlow
When I try to run RTs on Derecho using Rocoto or ecFlow, they hang. I first encountered this issue on PR #3107, but I had been hearing of people having similar issues before then. In those cases, it tended to happen when there were baseline changes and therefore a comparison error, but on #3107, there were no baseline changes. @jkbk2004 noticed that it sees to hang during or after the compile so that the tests don't actually run. 

Looking at the compile logs for, e.g., `compile_s2sw_intel`, I see: 
```
+ status='qstat: 5266962.desched1 Job has finished, use -x or -H to obtain historical job information'
++ awk '{print $5}'
+ status=finished,
+ case ${status} in
+ status_label=Unknown
+ echo 'rt_utils.sh: *** WARNING ***: Job status unsupported: finished,'
rt_utils.sh: *** WARNING ***: Job status unsupported: finished,
+ echo 'rt_utils.sh: *** WARNING ***: Status might be non-terminating, please manually stop if needed'
rt_utils.sh: *** WARNING ***: Status might be non-terminating, please manually stop if needed
+ echo '209 min. Pbs Job 5266962 Status: Unknown (finished,)'
209 min. Pbs Job 5266962 Status: Unknown (finished,)
+ ((  n=n+1  ))
+ wait 94586
+ sleep 60
```
It seems like the status given is "finished" but the one expected is "completed." 
My test directories are: 
* `/glade/derecho/scratch/gpetro/RTs/ufs-wm/3107a/tests` (w/Rocoto)
* `/glade/derecho/scratch/gpetro/RTs/ufs-wm/3107b/tests` (w/ecFlow)

My corresponding run directories are: 
* `/glade/derecho/scratch/gpetro/FV3_RT/rt_36809` (Rocoto)
* `/glade/derecho/scratch/gpetro/FV3_RT/rt_24003` (ecFlow)

### LINKED ISSUE #3290: default_vars.sh and tests/* are vulnerable to octal traps
## Description

An octal trap occurs when a variable with a leading zero (e.g., 00) is parsed by Bash as an octal (base-8) number rather than a decimal (base-10) number. For example,
```
SHOUR=00
export SECS=$(( SHOUR*3600 ))
```
Because SHOUR begins with 0, Bash treats it as an octal number. Since octal allows digits up to 7, the number remains 0. The math successfully evaluates to ``0 * 3600 = 0``. 

However, if ``SHOUR=08``, SHOUR will be evaluated as an invalid octal number and bash will throw a fatal error.

## Solution

Explicitly define the base to be used by prepending ``10#``

```
SHOUR=08
export SECS=$(( 10#SHOUR*3600 ))
```

DeniseWorthen:
## Commit Queue Requirements:

- [x] This PR addresses a relevant WM issue (if not, create an issue). 
- [x] All subcomponent pull requests (if any) have been reviewed by their code managers. NA
- [x] Run the full Intel+GNU RT suite (compared to current baselines), preferably on Ursa (Derecho or Hercules are acceptable alternatives). **Exceptions:** documentation-only PRs, CI-only PRs, etc.
   - [x] Commit log file w/full results from RT suite run (if applicable).
   - [x] Verify that `test_changes.list` indicates which tests, if any, are changed by this PR. Commit `test_changes.list`, even if it is empty.
- [X] Transparency in the use of generative AI is required by NOAA policy. Was GenAI used in this work? Yes
   - [X] [Generative AI tool (Gemini)] was used to assist with developing this code. The code has been reviewed, edited, and validated by NWS staff.

---

### #3289
Implements built-in bash for expressions in ``default_vars.sh`` and tests. Uses ``date`` to set time-related strings from variables. Ensures no-octal traps are present. Adds a custom GHA workflow to impose built-in bash and no octal-traps in the future. The GHA also contains a future check (currently commented out) to guard against the ``export VAR`` syntax which can be enabled when Issue #3287 is resolved.

### #3291 
Lately, RTs have been requiring 3-4 runs to get through the entire RT suite on Derecho. This fix allows the vast majority of tests to pass the first time. The fix was tested w/both ecFlow and Rocoto, and tests ran through without issue with both changes. 

### #3298 
A member of the CCPP team reached out to the EPIC US team with new links for the CCPP scientific documentation to be updated in different user guides.

### Commit Message:

```
* UFSWM - resolve inconsistent variable definitions in default_vars.sh and tests
* UFSWM - impose built-in bash for all arithmetic expressions; add date-based function to set date strings
* UFSWM - adds GHA to check for expr and octal traps 
* UFSWM - Derecho fix to allow RTs to run through in ~1 go
* UFSWM - update CCPP links
```

### Priority:

- [x] Normal

### UFSWM:

* Closes #3290 
* Partially resolves #3120 

* None

- [X] None

### Documentation:
- [X] Documentation update NOT required. 
   - Explanation: implements standard bash

---
- [X] No Baseline Changes.

### Input data Changes:

- [X] None.

### Library Changes/Upgrades:

- [X] No Updates
  
---

## Testing Log:
- RDHPCS
  - [x] Orion
  - [x] Hercules
  - [x] GaeaC6
  - [x] Derecho
  - [x] Ursa
- WCOSS2
  - [x] Dogwood/Cactus
  - [ ] Acorn
- [ ] CI

## Testing Remarks:

- Skipping Acorn testing due to system unavailability.


DusanJovic-NOAA:
I wonder if we really need to prefix every single variable in math expressions with 10#. Most variables aren't initialized with a leading zero, and is very unlikely they ever will be. For example why would anyone initialize CICE_NPROC with 0200. Adding '10#' everywhere reduces readability a lot, in my opinion.

Would it be acceptable to reduce the number of variables that actually need to be initialized with a leading zero, and only check the math expressions for those variables and ensure they are prefixed correctly?

tests/tests/cpld_control_c1152_v17
```diff
@@ -175,7 +175,7 @@ export DT_INNER=${DT_ATMOS}
 export DT_CICE=${DT_ATMOS}
 export DT_DYNAM_MOM6=900
 export DT_THERM_MOM6=1800
-export CICE_DIAGFREQ=`expr $((FHMAX-$FHROT)) \* 3600 / $DT_CICE`
+export CICE_DIAGFREQ=$(( $((10#$FHMAX-10#$FHROT)) * 3600 / 10#$DT_CICE ))
```
DusanJovic-NOAA: In expressions like this, I do not think the second `$((` inside the `$((` is necessary

DeniseWorthen:
This was my thinking---take CICE_NPROC as an example. It is used two ways, one to fill a namelist and one to calculate some other values which also go into the namelist. Fortran will happily accept 020. But when use to calculate in the scripts, we need to prevent a leading zero being mis-interpreted. OK, so we may need to deal w/ CICE_NPROC. But we need to then go through that process w/ each variable and decide whether it was safe or not to have a leading zero and then only prevent that use; but then we'd need a list of variables which were checked for in the GHA; and we'd need to potentially maintain that list of variables to account for new cases etc. 

My approach is very defensive in part to reduce the burden of individually identifying variables and to also account for user error (setting nproc=020).

DusanJovic-NOAA:
Can GHA be updated to first find all variables currently initialized with leading zero, and then check that only those are used in the math expressions with the '10#' prefix.

For example, on command line I think this will find all such variables:

```
$ grep -h -r '=0[0123456789]' default_vars.sh tests* | sed -e 's/[[:blank:]]*export //g' | sed -e 's/=.*//g' | sort | uniq
AOD_FRQ
CHOUR
FHMAX
FHROT
OCNRES
SDAY
SHOUR
SMONTH
```

Maybe something like this or an equivalent can be used in a GHA. Will this be reliable enough?

tests/tests/cpld_decomp_gfsv17
```diff
@@ -61,13 +61,13 @@ export FHMAX=12
...
 export OCN_tasks=$OCN_tasks_cpl_dcmp
 export ICE_tasks=$ICE_tasks_cpl_dcmp
 export WAV_tasks=$WAV_tasks_cpl_dcmp
 
 export CICE_NPROC=$ICE_tasks
-export np2=`expr $CICE_NPROC / 2`
-export CICE_BLCKX=`expr $NX_GLB / $np2`
-export CICE_BLCKY=`expr $NY_GLB / 2`
+export np2=$(( 10#$CICE_NPROC / 2 ))
+export CICE_BLCKX=$(( 10#$NX_GLB / 10#$np2 ))
```
DeniseWorthen: I don't think we should need the ``10#$np2`` here; since we're calculating it safely in the previous line.

On the other hand, NX_GLB and NY_GLB could be set to anything by a user; what if they give a value of "0360".

DeniseWorthen:
> Can GHA be updated to first find all variables currently initialized with leading zero, and then check that only those are used in the math expressions with the '10#' prefix.
> 
> For example, on command line I think this will find all such variables:
> 
> ```
> $ grep -h -r '=0[0123456789]' default_vars.sh tests* | sed -e 's/[[:blank:]]*export //g' | sed -e 's/=.*//g' | sort | uniq
> AOD_FRQ
> CHOUR
> FHMAX
> FHROT
> OCNRES
> SDAY
> SHOUR
> SMONTH
> ```
> 
> Maybe something like this or an equivalent can be used in a GHA. Will this be reliable enough?

So the logic would be something like "look through all assigned variables; if a variable is initialized with a leading-zero, check that it is not used in any expression which might trigger an octal-trap".

NickSzapiro-NOAA:
Is an alternative to use a "safer" calculator like [bc](https://www.gnu.org/software/bc/) where default is decimal ? Or python (?)

DusanJovic-NOAA:
> Is an alternative to use a "safer" calculator like [bc](https://www.gnu.org/software/bc/) where default is decimal ? Or python (?)

We could use bc, assuming it is available by default on all systems, which I believe it is, but the question remains, when and where will we use it? Everywhere, instead of all current uses of `$((...))`, or only in case one of the operands might be an octal number? How do we identify those?

DeniseWorthen:
My feeling is that we can only identify them by asking how they are used. That is, if a variable is assigned and then used to calculate or set another variable, then that use has to be octal-safe. 

Also, I did change all ``expr`` uses to be built-in bash. I did this primarily so that there would be a common way of calculating expressions but it does lead to including some lines that would not otherwise need to be included.

NickSzapiro-NOAA:
If there is a safe option that works, I would use it everywhere. Our AI friend suggests a wrapper like
```
safe_math() {
    # Usage: result=$(safe_math "08 + 1")
    #        result=$(safe_math "09 / 3.0")
    echo "$@" | bc -l
}
```

Edit: Or here is a version to strip leading zeros without bc
```
safe_math() {
    local expr
    
    # Use sed to strip leading zeros from any number in the string
    # For example: "08 + 009" becomes "8 + 9"
    expr=$(printf "%s" "$*" | sed 's/\(^\|[^0-9]\)00*\([0-9]\)/\1\2/g')
    
    # Evaluate using native Bash Base-10 arithmetic
    echo $(( expr ))
}
```

DeniseWorthen:
Can you show how that would read for this change

```
diff --git a/tests/tests/control_restart_p8_atmlnd b/tests/tests/control_restart_p8_atmlnd
index a86b1607..b4f3c499 100644
--- a/tests/tests/control_restart_p8_atmlnd
+++ b/tests/tests/control_restart_p8_atmlnd
@@ -29,8 +29,8 @@ export LIST_FILES="sfcf024.tile1.nc \
                    ufs.cpld.lnd.out.2021-03-23-21600.tile5.nc \
                    ufs.cpld.lnd.out.2021-03-23-21600.tile6.nc"

-export RESTART_FILE_PREFIX="${SYEAR}${SMONTH}${SDAY}.$(printf "%02d" $(( ${FHROT} + ${SHOUR} )))0000"
-export RESTART_FILE_SUFFIX_SECS="${SYEAR}-${SMONTH}-${SDAY}-$(printf "%02d" $(( (${FHROT}+${SHOUR})*3600 )))"
+export RESTART_FILE_PREFIX="${SYEAR}${SMONTH}${SDAY}.$(printf "%02d" $(( 10#${SHOUR} + 10#${FHROT} )))0000"
+export RESTART_FILE_SUFFIX_SECS="${SYEAR}-${SMONTH}-${SDAY}-$(printf "%02d" $(( (10#${FHROT}+10#${SHOUR})*3600 )))"
```

NickSzapiro-NOAA:
```
export RESTART_FILE_PREFIX="${SYEAR}${SMONTH}${SDAY}.$(printf "%02d" "$(safe_math "${FHROT} + ${SHOUR}")")0000"

export RESTART_FILE_SUFFIX_SECS="${SYEAR}-${SMONTH}-${SDAY}-$(printf "%02d" "$(safe_math "(${FHROT} + ${SHOUR}) * 3600")")"
```

DeniseWorthen:
Thanks. I guess I don't see that as "more readable" than the built-in bash 10#; but that is just my opinion; happy to implement that way if it is seen as preferable. 

How would you structure a GHA to enforce a "don't make any assignments that don't use safe_math" ?

NickSzapiro-NOAA:
I'm not sure if it's more readable in general ... personally, 10# looks very strange to me.

Maybe via a linter looking for math parentheses outside of safe_math. Untested:

```
#!/usr/bin/env bash
# lint_math.sh - Enforces the use of safe_math for Bash arithmetic

# 1. Define the safe_math wrapper so the linter can use it natively
safe_math() {
    local expr
    # Strip leading zeros to prevent the octal trap
    expr=$(printf "%s" "$*" | sed 's/\(^\|[^0-9]\)00*\([0-9]\)/\1\2/g')
    
    # Evaluate safely (using the inline exception so the linter ignores this line)
    echo $(( expr )) # nolint: safe_math
}

echo "Scanning for unsafe Bash arithmetic..."

TOTAL_VIOLATIONS=0
VIOLATION_LOG=""

# 2. Find all shell scripts, ignoring the .git directory
while IFS= read -r file; do
  
  FILE_VIOLATIONS=$(awk '
    # RULE 1: Ignore comment lines
    /^[[:space:]]*#/ { next }
    
    # RULE 2: Ignore lines explicitly bypassing the linter (like safe_math itself)
    /nolint: safe_math/ { next }
    
    # RULE 3: Ignore standard C-style for-loops
    /for[[:space:]]*\(\(/ { next }
    
    # RULE 4: Ignore math that explicitly uses the 10# base override
    /10#/ { next }
    
    # TRIGGER: Catch $(( ... )) OR standalone (( ... ))
    /\$\(\(/ || /(^|[[:space:]])\(\(/ {
        printf "❌ %s:%d: %s\n", FILENAME, FNR, $0
    }
  ' "$file")

  if [ -n "$FILE_VIOLATIONS" ]; then
    # wc -l often pads output with spaces (e.g., "       2") depending on the OS
    COUNT=$(echo "$FILE_VIOLATIONS" | wc -l)
    
    # 3. USE SAFE MATH to tally the violations! 
    # Because safe_math parses strings, it safely ignores wc's extra spaces.
    TOTAL_VIOLATIONS=$(safe_math "$TOTAL_VIOLATIONS + $COUNT")
    
    VIOLATION_LOG="${VIOLATION_LOG}${FILE_VIOLATIONS}"$'\n'
  fi

done < <(find . -type f -name "*.sh" -not -path "*/\.git/*")

# Final Enforcement Report
if [ "$TOTAL_VIOLATIONS" -gt 0 ]; then
  echo ""
  echo "🚨 ERROR: Found $TOTAL_VIOLATIONS unsafe raw Bash arithmetic violation(s)!"
  echo "--------------------------------------------------------"
  printf "%s" "$VIOLATION_LOG"
  echo "--------------------------------------------------------"
  echo "To prevent the octal trap (e.g., 08 being read as invalid base-8), please:"
  echo "  1. Use the 'safe_math' wrapper: \$(safe_math \"\$A + \$B\")"
  echo "  2. OR use the native Base-10 override: \$(( 10#\$A + 1 ))"
  echo "  3. OR append '# nolint: safe_math' to the line if you know it is safe."
  exit 1
else
  echo "✅ All math operations look safe! (0 violations)"
  exit 0
fi
```

DusanJovic-NOAA:
I think using `10#` is more explicit, it shows the intent more precisely, which in this context I find more useful. The term safe_math can mean different things, like preventing divide by zero or preventing overflow. It's not obvious when and where to use it. And I am not sure we can 'safely` use safe_math everywhere.

For example in rt.sh we have  code like:
```
      TPN=$(( TPN / THRD ))                                                                                                                                                     
      NODES=$(( TASKS / TPN ))
      if (( NODES * TPN < TASKS )); then
        NODES=$(( NODES + 1 ))
      fi
```
which assumes integer arithmetic (rounding), which I'm not sure how bc handles. I know this PR is only about default_vars.sh and tests/* file, but anyway.

If we can minimize the number of variables initialized with leading zero and use `10#` only in math expressions (`$(( .. ))`) and only for those variables then maybe that can be acceptable compromise.

For example, I see FHROT is almost always initialized as 'regular decimal number':

```
$ grep FHROT= default_vars.sh tests/*
default_vars.sh:    export FHROT=0
default_vars.sh:export FHROT=0
default_vars.sh:export FHROT=0
default_vars.sh:  export FHROT=0
default_vars.sh:  export FHROT=0
tests/control_c384gdas:export FHROT=3
tests/control_c384gdas_debug:export FHROT=3
tests/control_restart_noqr_p8:export FHROT=12
tests/control_restart_p8:export FHROT=12
tests/control_restart_qr:export FHROT=12
tests/control_stochy_restart:export FHROT=06
tests/control_wam:export FHROT=0
tests/control_wam_debug:export FHROT=0
tests/conus13km_restart:export FHROT=1
tests/conus13km_restart_mismatch:export FHROT=1
tests/cpld_control_c1152_v17:export FHROT=3
tests/cpld_control_gefs:export FHROT=0
tests/cpld_control_gfsv17_iau:export FHROT=3
tests/cpld_control_sfs:export FHROT=0
tests/cpld_restart_c1152_v17:export FHROT=18 #restart after any IAU
tests/cpld_restart_c12_9deg:export FHROT=3
tests/cpld_restart_c192_p8:export FHROT=15
tests/cpld_restart_c24_5deg:export FHROT=3
tests/cpld_restart_c24_9deg:export FHROT=3
tests/cpld_restart_c48_5deg:export FHROT=3
tests/cpld_restart_gefs:export FHROT=$((10#$FHROT+10#$FHROT_REST))
tests/cpld_restart_gfsv17:export FHROT=12
tests/cpld_restart_gfsv17_iau:export FHROT=18 #restart after any IAU
tests/cpld_restart_p8:export FHROT=12
tests/cpld_restart_sfs:export FHROT=12
tests/cpld_warmstart_c12_9deg:export FHROT=0
tests/cpld_warmstart_c24_5deg:export FHROT=0
tests/cpld_warmstart_c24_9deg:export FHROT=0
tests/cpld_warmstart_c48_5deg:export FHROT=0
tests/datm_cdeps_lm4_gswp3_rst:export FHROT=0
tests/datm_cdeps_lnd_era5_rst:export FHROT=0
tests/datm_cdeps_restart_cfsr:export FHROT=12
tests/gnv1_nested:export FHROT=0
tests/hrrr_control_restart:export FHROT=6
tests/hrrr_control_restart_dyn32_phy32:export FHROT=6
tests/hrrr_control_restart_noqr:export FHROT=6
tests/rap_restart:export FHROT=6
tests/rap_restart_dyn32_phy32:export FHROT=6
tests/rap_sfcdiff_restart:export FHROT=6
tests/regional_restart:export FHROT=3
```
only once as octal:
```
$ grep FHROT=0[01234567890] default_vars.sh tests/*
tests/control_stochy_restart:export FHROT=06
```

Which means if we fix this one instance, AND can be sure that it will never be initialized with leading zero (GHA can help) then we do not need to prefix each FHROT with 10#.

DusanJovic-NOAA:
I didn't see the second safe_math (the one without bc), the one that still uses `$((..))` internally. That's better than the one that uses the bc command.

DeniseWorthen:
> Which means if we fix this one instance, AND can be sure that it will never be initialized with leading zero (GHA can help) then we do not need to prefix each FHROT with 10#.

So I think this would mean for GHA keeping a list of variables which we can assert must not start with a leading 0. I agree that is possible as a starting place and probably pretty easy (a short list). But then we have to maintain that list. What if someone adds a new variable which they use in some expression. Wouldn't we have to then add that variable to the list that we check for non-0 leading digit?

EDIT: Maybe we don't need to maintain the list, it could be generated each time. "Identify any variable used in an expression; if it contains a leading 0 then error if not used in a octal-safe manner".

DeniseWorthen:
So I think we can implement a different strategy as suggested by Dusan. I created a test branch where I first did a straight replacement of built-in bash for expr. (I still think it good to stick to one method, and if we agree I could add a check against future expr use.)

Then I created a GHA that looked for any variable assignment with a leading zero which is then used in a calculation. Except for the FHROT=09 case Dusan found, all octal-traps are associated with SHOUR. I then created an octal-trap by initializing something w/ a NY_GLB w/ a leading zero in a test and it caught it.

I was stuck on the idea that we'd have to self-identify which variables would need to be tested, but I think that was wrong and Dusan had the better idea all along. I'll work on modifying the changes next week. I still want to poke the GHA lots of different ways to make sure I think it works.

DeniseWorthen:
@DusanJovic-NOAA I've updated the PR to reflect our email conversation. Thanks for your help on this.

I've added two local functions that use the ``date`` function to safely form the various RESTART file names as suggested. I've also removed the octal-safe changes for expressions which are not, under current assignment, vulnerable to octal-traps and removed zero-padded variable assignments where they were not actually needed. The ``SECS`` variable is now the only variable set w/ octal-safe ``10#${SHOUR} * 3600``.

The GHA now functions by finding any variable w/ a zero-pad and then determines whether it is a potential octal-trap risk. ``SHOUR`` is excluded because it is purposely set w/ zero-pad to feed in templates (eg ``cpld_control_run.IN``). For the remaining at-risk variables it checks whether it is used arithmetically and whether it is safe based on current assignment. So, for example, there are no false positives on CICE_BLCKX unless a user were to try to assign NX_GLB=0720. That would fail the GHA.

DusanJovic-NOAA:
> @DusanJovic-NOAA I've updated the PR to reflect our email conversation. Thanks for your help on this.
> 
> I've added two local functions that use the `date` function to safely form the various RESTART file names as suggested. I've also removed the octal-safe changes for expressions which are not, under current assignment, vulnerable to octal-traps and removed zero-padded variable assignments where they were not actually needed. The `SECS` variable is now the only variable set w/ octal-safe `10#${SHOUR} * 3600`.
> 
> The GHA now functions by finding any variable w/ a zero-pad and then determines whether it is a potential octal-trap risk. `SHOUR` is excluded because it is purposely set w/ zero-pad to feed in templates (eg `cpld_control_run.IN`). For the remaining at-risk variables it checks whether it is used arithmetically and whether it is safe based on current assignment. So, for example, there are no false positives on CICE_BLCKX unless a user were to try to assign NX_GLB=0720. That would fail the GHA.

There still few inconsistencies in expressions like:
```
export CICE_DIAGFREQ=$(( $((FHMAX-$FHROT)) * 3600 / $DT_CICE ))
```

Inner `$((` is not necessary. And if we use `$VAR` in arithmetic expressions everywhere, even though they are not required, then `FHMAX` should be `$FHMAX`. I personally prefer to not use `$` inside `$(( ... ))`.

DeniseWorthen:
Thanks. I'll try cleaning up the unnecessary ``$((``. I think the problem stemmed from how it was originally written as an expr using ``$((FHMAX-$FHROT))``, which I guess works but not clean.

BrianCurtis-NOAA:
> Thanks. I'll try cleaning up the unnecessary `$((`. I think the problem stemmed from how it was originally written as an expr using `$((FHMAX-$FHROT))`, which I guess works but not clean.

The linter should complain if there are any `$` inside the math `$((` calls, so that would be `$((FHMAX-FHROT))`

DeniseWorthen:
Then why didn't the linter complain? As far as I can tell, "superlinter" passed.

BrianCurtis-NOAA:
> Then why didn't the linter complain? As far as I can tell, "superlinter" passed.

I can't recall if we ignore remarks with the linter and only fix warnings, so it may pass. I think there's a remark when `$` are inside `$((`, could be wrong. but `$((FHMAX-FHROT))` is the proper way here

DeniseWorthen:
I ran it through shellcheck. Hopefully the syntax is standard now.

DusanJovic-NOAA (dismissed):
Looks good to me. Thanks for making all those changes and cleaning up the inconsistencies.

gspetro-NOAA:
@DeniseWorthen We're thinking of processing this PR next and combining in PR #3291 and PR #3298 . If you're good with that, feel free to combine in those two PRs (or tell me to do it, if you prefer). I'll update the description once everything's set.

DeniseWorthen:
Thanks, it should be good to go.

gspetro-NOAA:
Note that the failed log warnings/remarks check is just due to an older log (the last PR reduced warnings by 3 on a handful of tests).

FernandoAndrade-NOAA:
Acorn will be skipped for this PR due to the system being unavailable.

FernandoAndrade-NOAA (approved):

BrianCurtis-NOAA (approved):

gspetro-NOAA (approved):

PR #3190, opened 2026-04-10
Sync NCAR/main branch of CCPP physics with ufs/dev branch

grantfirl:
## Commit Queue Requirements:

- [ ] This PR addresses a relevant WM issue (if not, create an issue). 
- [x] All subcomponent pull requests (if any) have been reviewed by their code managers.
- [X] Run the full Intel+GNU RT suite (compared to current baselines), preferably on Ursa (Derecho or Hercules are acceptable alternatives). **Exceptions:** documentation-only PRs, CI-only PRs, etc.
   - [X] Commit log file w/full results from RT suite run (if applicable).
   - [X] Verify that `test_changes.list` indicates which tests, if any, are changed by this PR. Commit `test_changes.list`, even if it is empty.
- [X] Fill out all sections of this template.

---

## Description:
This is primarily just a CCPP physics update, although it contains some host-side metadata changes to work with the ccpp-physics changes. See https://github.com/ufs-community/ccpp-physics/pull/369 for a description of the CCPP-physics changes.

### Commit Message:

```
* UFSWM - Sync NCAR/main branch of CCPP physics with ufs/dev branch 
  * UFSATM - Sync NCAR/main branch of CCPP physics with ufs/dev branch
    * ccpp-physics - Sync NCAR/main branch of CCPP physics with ufs/dev branch
```

### Priority:

- [ ] Critical Bugfix: Reason
- [ ] High: Reason
- [X] Normal

### UFSWM:

* N/A

### Sub component Pull Requests:

* UFSATM: https://github.com/NOAA-EMC/ufsatm/pull/1092
  * ccpp-physics: https://github.com/ufs-community/ccpp-physics/pull/369

### UFSWM Blocking Dependencies:

- [ ] Blocked by #
- [X] None

### Documentation:

- [ ] Documentation update required. 
   - [ ] Relevant updates are included with this PR. 
   - [ ] A WM issue has been opened to track the need for a documentation update; a person responsible for submitting the update has been assigned to the issue (link issue).
- [X] Documentation update NOT required. 
   - Explanation: This just updates CCPP physics and doesn't change any results. It's possible that documentation changes should be done for the CCPP technical documentation related to the new `ccpp_bcast` capability.

---
### Regression Test Changes (Please commit test_changes.list):

- [ ] PR Adds New Tests/Baselines.
- [ ] PR Updates/Changes Baselines.
- [X] No Baseline Changes.

### Input data Changes:

- [X] None.
- [ ] PR adds input data.
[...]

PR #3097, opened 2026-02-17
Use cray wrappers for compilers/mpi on Derecho

natalie-perlin:
## Commit Queue Requirements:

- [x] This PR addresses a relevant WM issue (if not, create an issue). 
- [x] All subcomponent pull requests (if any) have been reviewed by their code managers.
- [x] Run the full Intel+GNU RT suite (compared to current baselines), preferably on Ursa 
   NB: Run a subset of 30 Intel+GNU RT suite on Derecho, not yet compared to current baselines.
(Derecho or Hercules are acceptable alternatives). **Exceptions:** documentation-only PRs, CI-only PRs, etc.
   - [x] Commit log file w/full results from RT suite run (if applicable).
   - [x] Verify that `test_changes.list` indicates which tests, if any, are changed by this PR. Commit `test_changes.list`, even if it is empty.
- [x] Fill out all sections of this template.

---

## Description:
Two issues are addressed in this PR:
1. Use cray wrappers for environmental variables FC, CC, CXX in modulefiles on Derecho, for both Intel and Gnu. They need to be specified as following:

setenv("CC","cc")
setenv("CXX","CC")
setenv("FC","ftn")

2. Corrected job status query for the complete jobs on Derecho/PBS scheduler, in ./tests/rt_utils.sh 

3.  Rebuilt spack-stack/1.9.2 environments with zlib/1.2.13 and glibc/2.38, resolving the misaligned library versions following system upgrade:
/glade/work/epicufsrt/contrib/spack-stack/derecho/spack-stack-1.9.2/envs/ue-oneapi-zlib.1.2.13/
/glade/work/epicufsrt/contrib/spack-stack/derecho/spack-stack-1.9.2/envs/ue-gcc-12.4-zlib-1.2.13/

NB: Due to changes in compilers and libraries used, the binaries differ. Thus, model output does not pass bit-to-bit comparison with the older baseline. New baseline tests are generated, and the full RT suite is successfully run against the newly created baseline. 

### Commit Message:
```
* UFSWM - Use cray wrappers for compilers/mpi on Derecho
* UFSWM - fix job status issue for PBS scheduler on Derecho
```

NEWER LOGS/ FULL RTs: 
Full regression test creating a baseline sumitted as:
` ./rt.sh -a NRAL0032 -r -c &> log.rt.out`
[RegressionTests_derecho_baseline.log](https://github.com/user-attachments/files/25605903/RegressionTests_derecho_baseline.log)
[...]

PR #3291, opened 2026-07-10
Derecho bugfix suggested in Issue #3120

### LINKED ISSUE #3120: Derecho hanging when using both Rocoto and ecFlow

gspetro-NOAA:
## Commit Queue Requirements:

- [x] This PR addresses a relevant WM issue (if not, create an issue). 
- [x] All subcomponent pull requests (if any) have been reviewed by their code managers.
- [x] Run the full Intel+GNU RT suite (compared to current baselines), preferably on Ursa (Derecho or Hercules are acceptable alternatives). **Exceptions:** documentation-only PRs, CI-only PRs, etc.
   - [x] Commit log file w/full results from RT suite run (if applicable).
   - [x] Verify that `test_changes.list` indicates which tests, if any, are changed by this PR. Commit `test_changes.list`, even if it is empty.
- [ ] Transparency in the use of generative AI is required by NOAA policy. Was GenAI used in this work? No. 
   - [ ] [Generative AI tool (insert name, if any)] was used to assist with developing this code. The code has been reviewed, edited, and validated by NWS staff. N/A

---

## Description:
Lately, RTs have been requiring 3-4 runs to get through the entire RT suite on Derecho. This fix will hopefully allow the vast majority of tests to pass the first time. 

### Commit Message:
```
* UFSWM - Derecho fix to allow RTs to run through in ~1 go
```

### Priority:
- [ ] Critical Bugfix: Reason
- [X] High: Ease Derecho testing burden for all PRs
- [ ] Normal

### UFSWM:
* Partially resolves #3120 

* None

- [ ] Blocked by #
- [X] None

### Documentation:
- [ ] Documentation update required. 
   - [ ] Relevant updates are included with this PR. 
   - [ ] A WM issue has been opened to track the need for a documentation update; a person responsible for submitting the update has been assigned to the issue (link issue).
- [X] Documentation update NOT required. 
   - Explanation: Only touches RT testing internals

---
- [ ] PR Adds New Tests/Baselines.
- [ ] PR Updates/Changes Baselines.
- [X] No Baseline Changes.

### Input data Changes:
- [X] None.
- [ ] PR adds input data.
- [ ] PR changes existing input data.

### Library Changes/Upgrades:
- [ ] Required
  * Library names w/versions:
  * Git Stack Issue (JCSDA/spack-stack#)
- [X] No Updates
  
---

## Testing Log:
- RDHPCS
  - [ ] Orion
  - [ ] Hercules
  - [ ] GaeaC6
  - [ ] Derecho
  - [ ] Ursa
- WCOSS2
  - [ ] Dogwood/Cactus
  - [ ] Acorn
- [ ] CI

## Testing Remarks:

-


DeniseWorthen:
@gspetro-NOAA FYI, I did have to wait quite a long time in the Q the first time I ran; but I also noticed that 10% of Derecho's nodes were reserved at that point. My second run (w/o any active node reservation) went through much faster.

gspetro-NOAA:
> @gspetro-NOAA FYI, I did have to wait quite a long time in the Q the first time I ran; but I also noticed that 10% of Derecho's nodes were reserved at that point. My second run (w/o any active node reservation) went through much faster.

Yes, I think this change resolves one part of the Derecho issue--the need to run 3-4 times to get the full test set to pass. I think there are larger issues on Derecho though. The RTs _can_ run in under 3 hours, but I have seen them take as much as 18, which is really impossible for keeping any kind of cadence/schedule to PR processing. And RTs on Derecho take about twice as long as other RDHPCS. So I think there may be additional issues at play. Getting 98%+ of RTs to pass in a single run would be huge though.

gspetro-NOAA:
More Derecho problems... I'm getting `PBS: job killed: walltime 3635 exceeded limit 3600` type messages in a bunch of err files... including for the test I was running on PR 3174, so it doesn't seem specific to this PR... thus far, I've received 7 "exceeded wall clock time" emails from root for this PR.

DeniseWorthen:
Do you get system emails; I just got one

```
Dear NSF NCAR HPC Users,

CISL services are degraded due to an issue on the Destor filesystem impacting all clients.  New logins likely delayed; and some running jobs may time out due to queue walltime exceedences.  System engineers are restoring the services for a planned restart before an estimated 11am; but successful completion of ongoing work will dictate return to service.

Thank you for your patience and understanding.

START: Fri Jul 10 2026 10:15 AM MDT
DURATION: 1 hour

SERVICES AFFECTED
CISL Status
```

I read this as "Derecho filesystem" but maybe there is actually something called "Destor".

gspetro-NOAA:
> Do you get system emails; I just got one
> 
> ```
> Dear NSF NCAR HPC Users,
> 
> CISL services are degraded due to an issue on the Destor filesystem impacting all clients.  New logins likely delayed; and some running jobs may time out due to queue walltime exceedences.  System engineers are restoring the services for a planned restart before an estimated 11am; but successful completion of ongoing work will dictate return to service.
> 
> Thank you for your patience and understanding.
> 
> START: Fri Jul 10 2026 10:15 AM MDT
> DURATION: 1 hour
> 
> SERVICES AFFECTED
> CISL Status
> ```
> 
> I read this as "Derecho filesystem" but maybe there is actually something called "Destor".

I was just coming back here to post a "never mind!" message bc I just saw the email. 😆 🙄

DeniseWorthen:
@gspetro-NOAA Looks like it worked?

gspetro-NOAA:
> @gspetro-NOAA Looks like it worked?

It did! However, we're putting a hold on the PR for now because ultimately, this is primarily a timing change that is probably papering over a bigger issue. 🙁 I'm not positive the change in line 320 is doing much, since without the sleep 160 change, we were still having the issues running RTs before Natalie's Derecho PR (#3097), which changed line 320 to its latest form in develop. In fact, it looks like the RT problems began with PR #3190, which went in on May 11, over a month before the Derecho-specific PR.

MichaelLueken:
Hi @gspetro-NOAA and @DeniseWorthen -

Figured I would chime in with some information regarding the Rocoto issue on Derecho.

One of the major issues on Derecho is that `qstat` is no longer`qstat`.  The system administrators for Derecho replaced `qstat` with their own [qstat-cache](https://github.com/NCAR/qstat-cache).  Chris Harrop [identified](https://github.com/NOAA-GSL/rocoto/issues/121#issuecomment-4270946617) this back on April 17.  If a job id hasn't been cached before the sleep timer is up, then the job will be identified as `UNKNOWN` in `Rocoto`, and the job will fail.  Increasing the sleep timer, as done on line [328](https://github.com/ufs-community/ufs-weather-model/pull/3291/changes#diff-f127dc09517d17232da090b4386e4246fbfa8798be0fc927bb40a0330991f20bR328), ensures that the job id gets cached, allowing the test to successfully pass.

gspetro-NOAA:
> Hi @gspetro-NOAA and @DeniseWorthen -
> 
> Figured I would chime in with some information regarding the Rocoto issue on Derecho.
> 
> One of the major issues on Derecho is that `qstat` is no longer`qstat`. The system administrators for Derecho replaced `qstat` with their own [qstat-cache](https://github.com/NCAR/qstat-cache). Chris Harrop [identified](https://github.com/NOAA-GSL/rocoto/issues/121#issuecomment-4270946617) this back on April 17. If a job id hasn't been cached before the sleep timer is up, then the job will be identified as `UNKNOWN` in `Rocoto`, and the job will fail. Increasing the sleep timer, as done on line [328](https://github.com/ufs-community/ufs-weather-model/pull/3291/changes#diff-f127dc09517d17232da090b4386e4246fbfa8798be0fc927bb40a0330991f20bR328), ensures that the job id gets cached, allowing the test to successfully pass.

@MichaelLueken Thanks for these details! It looks like this may be behind the failures with both ecFlow and Rocoto. 

Some questions for those w/thoughts/opinions and more knowledge than me on the matter: 
- Could we manually bypass to use native qstat? (And any ideas on how? Pros/cons?) 
- What is the best path forward? Contact sys admins? File an issue in qstat-cache it the hope that they can speed up the job-ID caching? Or at least tell us the cache refresh rate? Other suggestions?
   - Copilot's look at the repo says "The cache refresh rate in qstat-cache is configured via  site.cfg 's  [cache]  section... So effectively data is refreshed as often as every `Frequency` seconds (default 10s in the example config), driven by a per-minute cron calling  `util/gen_data.sh` / `gen_data_remote`." I suspect the example config and Derecho's rate are significantly different and/or the cron tab update is slow. 
- Is there any harm in increasing the sleep time to accommodate what we know works for qstat-cache on Derecho? Or should we be requesting improvements to qstat-cache...? 
- How does this issue intersect with the long queue time issue (if at all)? 

@christopherwharrop-noaa Feel free to chime in if you have thoughts!

MichaelLueken:
@gspetro-NOAA -

* `Could we manually bypass to use native qstat? (And any ideas on how? Pros/cons?)`
No, native qstat no longer exists on Derecho, only the qstat-cache.

* `What is the best path forward? Contact sys admins? File an issue in qstat-cache it the hope that they can speed up the job-ID caching? Or at least tell us the cache refresh rate? Other suggestions?`
Increasing the sleep time is the best path forward.  This has been an issue since the Derecho system maintenance on July 25, 2025, so close to a year now.  Several tickets have been opened and nothing has been done.  To continue to support Derecho, we need to create our own fix to this, which is increasing the sleep time.

* `Is there any harm in increasing the sleep time to accommodate what we know works for qstat-cache on Derecho? Or should we be requesting improvements to qstat-cache...?`
No, there is no harm increasing the sleep time.  The SRW App has been running with a 2 and a half minute (150 second) sleep time on Derecho for the last year without issue.

* `How does this issue intersect with the long queue time issue (if at all)?`
It doesn't intersect with the long queue time issue.  The long queue time issue is due to jobs being part of some large efforts and NSF Directorate allocations with queue wallclock exceptions applied after being requested and validated as needed.  There is nothing we can really do, unfortunately.  If we had an actual account (not the director's NRAL0032 account), then we might have some priority, but that's not the case.

christopherwharrop-noaa:
I am exploring a more direct approach from the Rocoto side, motivated by issues on both Derecho and WCOSS.  The basic problem is that PBSPro has poor multithreading capability and is easily overwhelmed when users issue too many qstat calls per unit time.  It is compounded by the need for the -xf options which allow extraction of exit status information from completed jobs. So administrators have resorted to caching qstat output to reduce load on the PBS services.  This means the qstat information is old and how old depends on the frequency of the caching that has been implemented.  But it also means that administrators have full control over the qstat load.

A much better approach exists, but it has to be implemented inside Rocoto.  PBS has an API (via libpbs.so) that allows you to query the PBS service directly from code.  And the API allows you to request only the specific information you really need.  A big part of the problem with the qstat command is that it is very heavy handed and there are no options for requesting small amounts of job information that also includes exit status for completed jobs.  Rocoto is forced to run one of the most expensive commands against the service because there simply isn't any other option.  Except for the API.  I have already done a proof of concept implementation in a Rocoto branch to demonstrate calling the API, and it works. And it's also quite fast.  It's mostly a question of carving out the time to harden it from prototype to production status and then releasing it.  Folks who use WCOSS are really wanting this also.

gspetro-NOAA:
> @gspetro-NOAA -
> 
> * `Could we manually bypass to use native qstat? (And any ideas on how? Pros/cons?)`
>   No, native qstat no longer exists on Derecho, only the qstat-cache.
> * `What is the best path forward? Contact sys admins? File an issue in qstat-cache it the hope that they can speed up the job-ID caching? Or at least tell us the cache refresh rate? Other suggestions?`
>   Increasing the sleep time is the best path forward.  This has been an issue since the Derecho system maintenance on July 25, 2025, so close to a year now.  Several tickets have been opened and nothing has been done.  To continue to support Derecho, we need to create our own fix to this, which is increasing the sleep time.
> * `Is there any harm in increasing the sleep time to accommodate what we know works for qstat-cache on Derecho? Or should we be requesting improvements to qstat-cache...?`
>   No, there is no harm increasing the sleep time.  The SRW App has been running with a 2 and a half minute (150 second) sleep time on Derecho for the last year without issue.
> * `How does this issue intersect with the long queue time issue (if at all)?`
>   It doesn't intersect with the long queue time issue.  The long queue time issue is due to jobs being part of some large efforts and NSF Directorate allocations with queue wallclock exceptions applied after being requested and validated as needed.  There is nothing we can really do, unfortunately.  If we had an actual account (not the director's NRAL0032 account), then we might have some priority, but that's not the case.

FYI @BrianCurtis-NOAA @dpsarmie @FernandoAndrade-NOAA @jkbk2004 @NickSzapiro-NOAA 
It looks like moving forward with PR #3291 is likely the best path forward to accommodate either job scheduler on Derecho. 
And also, it looks like we need management to push for an actual EPIC account or priority queue (or both). Otherwise, PR processing will remain at the mercy of whoever else is running jobs on Derecho, which is not sustainable. I suspect there are restrictions re: needing NSF funding for an account, but I would think there would be a workaround for that if teams across NOAA and NCAR really want us supporting Derecho.

tests/rt_utils.sh
```diff
@@ -315,17 +315,17 @@ submit_and_wait() {
       pbs)
         set +e
         job_info=$( qstat "${jobid}" )
-	job_info=$( grep -v "Job has finished, use -x or -H to obtain historical job information" <<< "${job_info}")
+        job_info=$( grep -v "Job has finished, use -x or -H to obtain historical job information" <<< "${job_info}")
         set -e
-        if [[ ${job_info} == *"${jobid}"* && ${status_label:-} != "Unknown" ]]; then
+        if grep -q "${jobid}" <<< "${job_info}"; then
```
gspetro-NOAA, line 320: @DeniseWorthen My only question is why you decided to change this line? I ran RTs successfully w/only the sleep 160 change on Jiande's PR yesterday, so this doesn't seem totally necessary.

DeniseWorthen: I was getting odd failures (something about "unknown") on a bunch of rap tests and some others. I could see that the actual rt_testname.log file showed them as passing, but they were getting flagged as failing.

Also, the original comment from in the original issue basically said "just change the qstat time to 150" and since the change on LN 320 had apparently not helped (you were still having problems on Derecho since the change), I didn't think it was the actual solution.

gspetro-NOAA: Oh, interesting... 🤔 Were you getting those "Unknown" failures even after changing to `sleep 160`? It seems that the Unknown status is returned "if a job id hasn't been cached before the sleep timer is up". But if you had already changed to `sleep 160` then maybe we should revert rt_utils.sh.
@natalie-perlin Do you remember why you made this change in #3097?

gspetro-NOAA:
One remaining point, just to cover my bases: Why did this problem of several reruns not start popping up until PR #3190 ? I think (based on checking the five Derecho logs previous to PR 3190) that PR 3190 is when I switched over to running with ecFlow, which I did after having Rocoto hang on me one too many times, possibly due to the same underlying qstat/caching issue. 🤔 So it seems that I traded one problem for another. Hopefully this PR can resolve the issue for the WM on Derecho.

tests/rt_utils.sh
DeniseWorthen: The first thing I tried was just increasing the sleep time, quickly got weird behavior and then decided to revert the jobid query also. I know those attempts (one aborted, one successful) were during the period Derecho had a 10% node reservation. I didn't try a case w/ just the jobid query reverted.  With both changes, and the next day w/ no node reservation, everything ran straight through.

natalie-perlin: @gspetro-NOAA  - yes I do remember. It was needed because otherwise it would report a job as "failed", even it was still processing. I don't remember more details, but yes, it was a "fix" specifically for Derecho.

natalie-perlin: Rocoto relies on decoding job state from job manager, and there was/is apparently some mismatch between newer/older versions of recognizing the job state. I think with Derecho, if the job has finished it would not be accessible with qsub without extra flags and extra decoding.

gspetro-NOAA: Ok, I tested with and without this change, and without it, I'm seeing the errors Denise mentioned related to "Unknown" on two tests, so it seems that with ecFlow, at least, maybe the original syntax is better: 

```
echo 'rt_utils.sh Job (6745997) is now in the queue.'
rt_utils.sh Job (6745997) is now in the queue.
+ local n=1
+ [[ true == \f\a\l\s\e ]]
+ case ${SCHEDULER} in
+ set +e
++ qstat 6745997
+ job_info='qstat: Unknown Job Id 6745997.desched1'
++ grep -v 'Job has finished, use -x or -H to obtain historical job information'
+ job_info='qstat: Unknown Job Id 6745997.desched1'
+ set -e
+ [[ qstat: Unknown Job Id 6745997.desched1 == *\6\7\4\5\9\9\7* ]]
+ [[ '' != \U\n\k\n\o\w\n ]]
+ job_running=true
++ grep 6745997
+ status='qstat: Unknown Job Id 6745997.desched1'
++ awk '{print $5}'
+ status=6745997.desched1
+ case ${status} in
+ status_label=Unknown
+ echo 'rt_utils.sh: *** WARNING ***: Job status unsupported: 6745997.desched1'
rt_utils.sh: *** WARNING ***: Job status unsupported: 6745997.desched1
+ echo 'rt_utils.sh: *** WARNING ***: Status might be non-terminating, please manually stop if needed'
rt_utils.sh: *** WARNING ***: Status might be non-terminating, please manually stop if needed
+ echo '1 min. Pbs Job 6745997 Status: Unknown (6745997.desched1)'
1 min. Pbs Job 6745997 Status: Unknown (6745997.desched1)
+ ((  n=n+1  ))
+ wait 121904
+ sleep 60
+ [[ true == \f\a\l\s\e ]]
+ case ${SCHEDULER} in
+ set +e
++ qstat 6745997
+ job_info='Job id            Name             User             Time Use S Queue           
----------------- ---------------- ---------------- -------- - -----
6745997.desched1  run_control_p8_* gpetro                --  Q cpu             '
++ grep -v 'Job has finished, use -x or -H to obtain historical job information'
+ job_info='Job id            Name             User             Time Use S Queue           
----------------- ---------------- ---------------- -------- - -----
6745997.desched1  run_control_p8_* gpetro                --  Q cpu             '
+ set -e
+ [[ Job id            Name             User             Time Use S Queue           
----------------- ---------------- ---------------- -------- - -----
6745997.desched1  run_control_p8_* gpetro                --  Q cpu              == *\6\7\4\5\9\9\7* ]]
+ [[ Unknown != \U\n\k\n\o\w\n ]]
+ job_running=false
+ status=COMPLETED
+ sleep 160
+ set +e
++ qstat 6745997 -x -f
++ grep Exit_status
++ awk '{print $3}'
+ exit_status=
+ set -e
+ [[ '' != 0 ]]
+ status=FAILED
+ case ${status} in
+ echo 'rt_utils.sh: !!!!!!!!!!JOB TERMINATED!!!!!!!!!! status=FAILED'
rt_utils.sh: !!!!!!!!!!JOB TERMINATED!!!!!!!!!! status=FAILED
```
See, e.g., `/glade/derecho/scratch/gpetro/RTs/ufs-wm/3078/tests/logs/log_derecho/run_control_p8_rrtmgp_rad32_intel.log` and `/glade/derecho/scratch/gpetro/RTs/ufs-wm/3078/tests/logs/log_derecho/run_rrfs_v1nssl_intel.log`.

DeniseWorthen: That does look like the message I was seeing.

gspetro-NOAA: Ok, I ran with and without this change w/Rocoto, and it worked without issue both ways. Whereas ecFlow seems to need it to prevent those "Unknown" errors. 

Logs w/Rocoto & proposed change: `/glade/derecho/scratch/gpetro/RTs/ufs-wm/pretests/3291/tests/logs/RegressionTests_derecho.log`
Logs w/Rocoto & w/o proposed change: `/glade/derecho/scratch/gpetro/RTs/ufs-wm/pretests/3291r-nofix/tests/logs/RegressionTests_derecho.log`

gspetro-NOAA:
Merged via #3289 .

